PR #7, opened 2020-03-27
Add new cleanup operator resources playbook

scoheb:


scoheb:
tested locally with `crc`

scoheb:
when using these roles locally thru `local-test-operator.yml`, it was called. 
this PR enables us to use the role thru a new playbook.

PR #14, opened 2020-04-24
Remove namespace from pull secret

sonam1412:
namespace removed when applying the secrets to avoid conflicts
Null case handled
Image source test UBI is now running on 4.x cluster

### DIFF
--- a/roles/inject_openshift_kube_objects/tasks/main.yml
+++ b/roles/inject_openshift_kube_objects/tasks/main.yml
@@ -15,16 +15,23 @@
   
 - name: "Decrypt the kube_objects through the shared symmetric key"
   shell: "openssl aes-256-cbc -d -in {{ kube_objects_dir }}/get_kubeObjects.txt -out {{ kube_objects_dir }}/kube_objects.yaml -pass file:{{ kube_objects_dir }}/secret.key"
-     
+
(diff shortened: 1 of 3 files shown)


sonam1412:
@dirgim @scoheb @lslebodn @nmars please review

roles/inject_openshift_kube_objects/tasks/main.yml
```diff
@@ -15,16 +15,23 @@
...
+
+- name: "Read yaml file content to get the namespace"
+  set_fact: 
+    content="{{ lookup('file', '{{ kube_objects_dir }}/kube_objects.yaml') | from_yaml }}"
+  no_log: true
+  ignore_errors: true   
   
 - name: "Verify kube_objects created on openshift cluster"
-  shell: "oc get secrets -n {{ openshift_namespace }}"
+  shell: "oc get secrets -n {{ content.metadata.namespace }}"
```
scoheb, line 31: I guess this implies we are only expecting secrets in the kube_objects...for now this is true...but in time it is possible that we will not have any right?

sonam1412: Yes we may have different kube resources or none. This is as per the current requirement just to make sure it is created and pipeline will not fail if there are any issues in this step

PR #18, opened 2020-05-13
Add operator bundle basic sanity tests to the parsing role

dirgim:
Operator bundle images have their basic info stored in two places:

* Bundle image labels
* The metadata/annotations.yaml file on the bundle image

Since this information is read from different places by diffrerent components, it is crucial that this data is exactly equal in both locations.

Move the OCP versions validation to the bundle sanity checks.

### DIFF
--- a/roles/parse_operator_bundle/tasks/bundle_sanity_checks.yml
+++ b/roles/parse_operator_bundle/tasks/bundle_sanity_checks.yml
@@ -1,4 +1,10 @@
 ---
+# Note: we probably want to make this more robust in the future to make a sliding scale that can be adjusted on the fly based on configuration?
+- name: "Verify ocp versions are valid"
+  fail:
+    msg: "com.redhat.openshift.versions must only contain v4.5"
+  when: (ocp_versions|length==0) or (ocp_versions|length>1) or ocp_versions[0] != "v4.5"
+
 - name: "Read the variables from annotations.yaml"
   include_vars:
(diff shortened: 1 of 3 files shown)


roles/parse_operator_bundle/tasks/bundle_sanity_checks.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- debug:
+    var: skopeo_inspect_json.Labels
+
+- debug:
+    var: annotations_vars.annotations
+
+- name: "Check if the annotation.yaml variables match the bundle image labels"
+  fail:
+    msg: "Values in the annotations yaml don't match the bundle image labels"
```
lslebodn, line 15: Will ansible tell us which condition does not match? or we will see just debug message.
It would be good to have useful error reporting in case of failure.

dirgim: There are debug messages displaying the fields that will be tested prior to making the check, which should make it obvious which fields don't match. I'm open to splitting this into separate checks though.

lslebodn: Sure they are there but it would be simpler to find a bug if we at least log "key" (e.g. `operators.operatorframework.io.bundle.channel.default.v1`) which is problematic/does not match

dirgim: Ok, changing

PR #19, opened 2020-05-15
Modify base64 decode command to fix the "bad magic number" error

sonam1412:
Fix the "bad magic number" error when decrypting the kube_objects
Fail if the shared symmetric key is not correct

### DIFF
--- a/roles/inject_openshift_kube_objects/tasks/main.yml
+++ b/roles/inject_openshift_kube_objects/tasks/main.yml
@@ -3,14 +3,23 @@
   shell: "mkdir -p {{ kube_objects_dir }}"
 
 - name: "Decode the symmetric key received from the user"
-  shell: "echo \"{{ symmetric_key }}\" | openssl enc -d -base64 > {{ kube_objects_dir }}/get_sym_key"
+  shell: "echo \"{{ symmetric_key }}\" | base64 --decode > {{ kube_objects_dir }}/get_sym_key"
   no_log: true
   
 - name: "Decrypt the symmetric key using RSA private key"
   shell: "openssl rsautl -decrypt -oaep -inkey {{ rsa_private_key }} -in {{ kube_objects_dir }}/get_sym_key -out {{ kube_objects_dir }}/secret.key"
 
+- name: "Verify the shared symmetric key"
+  shell: "stat -c %s {{ kube_objects_dir }}/secret.key"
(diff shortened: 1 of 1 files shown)


roles/inject_openshift_kube_objects/tasks/main.yml
```diff
@@ -3,14 +3,23 @@
...
+  register: symmetric_key_size
+
+- name: "Fail if the shared symmetric key is not 32 bytes"
+  fail:
+    msg: "Failed: The shared symmetric key should be of 32 bytes"
+  when: symmetric_key_size.stdout != "32"
+
 - name: "Read the encrypted kube_objects decode and store them in a file"
-  shell: "echo \"{{ kube_objects }}\" | tr \" \" \"\n\" | openssl enc -d -base64 > {{ kube_objects_dir }}/get_kubeObjects.txt"
+  shell: "echo \"{{ kube_objects }}\" | tr \" \" \"\n\" | base64 --decode > {{ kube_objects_dir }}/get_kubeObjects.txt"
```
lslebodn, line 22: Not sure whether it is the same issue but I filed a bug for openssl
https://bugzilla.redhat.com/show_bug.cgi?id=1836122

`base64` seems to be more robust
+1

roles/inject_openshift_kube_objects/tasks/main.yml
```diff
@@ -3,14 +3,23 @@
...
 - name: "Decrypt the symmetric key using RSA private key"
   shell: "openssl rsautl -decrypt -oaep -inkey {{ rsa_private_key }} -in {{ kube_objects_dir }}/get_sym_key -out {{ kube_objects_dir }}/secret.key"
 
+- name: "Verify the shared symmetric key"
+  shell: "stat -c %s {{ kube_objects_dir }}/secret.key"
+  register: symmetric_key_size
+
+- name: "Fail if the shared symmetric key is not 32 bytes"
+  fail:
+    msg: "Failed: The shared symmetric key should be of 32 bytes"
```
lslebodn, line 18: can we add  "symmetric_key_size" to the debug message.
Would be good for debugging purposes.

roles/inject_openshift_kube_objects/tasks/main.yml
sonam1412: quite similar

PR #20, opened 2020-05-22
[CVP-970] [CVP-1071] Improve output of prepare-metadata step for users

samvarankashyap:
Currently according to the upstream playbook the following happens:
1. Playbook takes two inputs that can be overriden by
   operator_dir: directory consisting of unzipped operator metadata zip
   operator_work_dir: directory which is usually the empty directory which
                      in which the operator metadata is copied and parsed
2. Given both attributes, operator data is being parsed and outputs to 
   and simple jinja template based file. 
   ```
   package_name: operatorname
   package_path: data2_work_dir/operatorname.package.yaml
   current_channel: stable
   current_csv: operatorname.vx.x.x
   current_csv_dir: data2_work_dir/x.x.x
   csv_path: data2_work_dir/x.x.x/operatorname.vx.x.x.clusterserviceversion.yaml
   operator_pod_name: xxxx-operator
   operator_capabilities: Seamless Upgrades
   operator_container_name: xxx-operator
   operator_allnamespaces_support: True
   operator_ownnamespace_support: False
   operator_singlenamespace_support: False
   operator_multinamespace_support: False
   crd_paths:
   - data2_work_dir/x.x.x/operatorname.crd.yaml
   - data2_work_dir/x.x.x/operatorname.crd.yaml
   - data2_work_dir/x.x.x/operatorname.crd.yaml

   ```

Observations:
Through out the playbook shell commands have been extensively used.  
Further, binaries like
yq, jq have also been used to fetch specific attributes from JSON documents. 
This is effecting the readablity and ease of debugging steps inside the 

Approach:
Improvements can be made by writing pure ansible code to do the above tasks 
with built in functionality of Ansible. Which increases the readability. 

Further, A simple block level exception handling needs to implemented 
to know where the ansible task has failed and what is the result of the failure. 
Which also acts as debugger whenever there is a specific failure inside the
the playbook run

The following PR has been created to resolve the issues to update the playbooks 
and add few debug statements and exception statements using block and rescue 
constructs inside Ansible.

As a positive side effect of this PR CVP-1071 is also resolved. 
ansible playbook fails on parse-operator-metadata fails on
special characters like '+' being included inside the operator

How to test this PR: 
1. Download metadata.zip file of operator
2. Unzip the contents of metadata.zip into folder say data_dir
3. Create another directory data_work_dir
4. Create another directory work_dir to collect output of parsed file
5. Run ansible prepare-operator-metadata.yml
```
ansible-playbook -vvvv prepare-operator-metadata.yml -e"operator_dir=./data_dir/"
 -e"operator_work_dir=./data_work_dir/" -e"work_dir=./work_dir/"parsed_operator_data.yml
```
6. The you should be able to get output of parsed operator-metadata as output in work_dir in parsed_operator_data.yml

### DIFF
--- a/roles/parse_operator_metadata/tasks/main.yml
+++ b/roles/parse_operator_metadata/tasks/main.yml
@@ -1,109 +1,163 @@
----
-- name: "Determine the package path in the operator metadata directory"
-  shell: "egrep -l \"packageName\" {{ operator_work_dir }}/* | head -n 1"
-  register: package_path_result
+- name: "Running Prepare operator metadata"
+  block:
 
-- set_fact:
-    package_path: "{{ package_path_result.stdout }}"
+  - name: "Determine the package path inside operator metadata directory"
+    find:
+      paths: "{{ operator_work_dir }}"
+      file_type: file
+      contains: 'packageName'
+    register: package_path_result 
 
-- name: "Determine the package name for the operator"
-  shell: "{{ yq_bin_path }} r {{ package_path }} 'packageName'"
-  register: package_name_result
+    # assuming only one file matches the file path
+  - name: "set variable for package_path from package_path_result"
+    set_fact:
+      package_path: "{{ package_path_result.files[0].path }}"
 
-- set_fact:
-    package_name: "{{ package_name_result.stdout }}"
+  - name: "Load variables from package path into variable package_vars"
+    include_vars:
+      file: "{{ package_path }}"
+      name: package_vars
 
-- debug:
-    var: current_channel
+  - name: "GET package_name variable from loaded package_vars"
+    set_fact:
+      package_name: "{{ package_vars['packageName'] }}"
 
-- name: "Try to get the operator's defaultChannel if exists"
-  shell: "{{ yq_bin_path }} r {{ package_path }} 'defaultChannel'"
-  register: default_channel_result
-  when: current_channel == ''
+  # Note:
+  # current logic will check the length of current_channel
+  # if the current_channel is empty string
+  # it defaults to defaultChannel attribute inside package_vars
+  # if the defaultChannel is undefined it fails over to
+  # first channel name inside package_var channels
+  - name: "Set default channel when the current_channel is empty"
+    set_fact:
+      current_channel: "{{ (current_channel |length > 0) |
+                            ternary(current_channel,
+                                    package_vars['defaultChannel'] |
+                                    default(package_vars['channels'][0]['name'])) }}"
 
-- set_fact:
-    current_channel: "{{ default_channel_result.stdout}}"
-  when: current_channel == ''
+  # Note:
+  # the above logic do not cover the case when
+  # the defaultChannel is empty string (It covers when defaultChannel is undefined)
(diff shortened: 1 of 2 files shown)


roles/parse_operator_metadata/tasks/main.yml
```diff
@@ -1,109 +1,151 @@
...
+
+    - name: "Rescue block contains the error messages"
+      debug:
+        msg: "We caught an error, can do stuff here to fix it, :-)"
+
+    - name: "FAILED task name in ansible is as follows:"
+      debug:
+        msg: "{{ ansible_failed_task.name }}"
+
+    - name: "FAILED task name in ansible is as follows:"
```
scoheb, line 141: duplicated?

roles/parse_operator_metadata/tasks/main.yml
```diff
@@ -1,109 +1,151 @@
...
+  - name: "Output all collected data to a yaml file in work dir"
+    template:
+      src: "parsed_operator_data.yml.j2"
+      dest: "{{ work_dir }}/parsed_operator_data.yml"
+
+  rescue:
+
+    - name: "Rescue block contains the error messages"
+      debug:
+        msg: "We caught an error, can do stuff here to fix it, :-)"
```
scoheb, line 135: Are we expecting to improve this in this PR?

roles/parse_operator_metadata/tasks/main.yml
samvarankashyap: @scoheb Yes i will update it

roles/parse_operator_metadata/tasks/main.yml
samvarankashyap: @scoheb : The rescue block contains all the debugs statments required, 
including which task has failed. 
what is the failed task output.
Inorder to understand easily i have added specific names to all tasks for end users to understand what task has failed.

samvarankashyap: @scoheb Are they any other details missing ?

samvarankashyap:
@scoheb PTAL

samvarankashyap:
@scoheb Tests have passed, 
I have updated the debug task value too.

scoheb:
I think @dirgim should review this too

roles/parse_operator_metadata/tasks/main.yml
```diff
@@ -1,109 +1,147 @@
...
+
+  - name: "GET package_name variable from loaded package_vars"
+    set_fact:
+      package_name: "{{ package_vars['packageName'] }}"
+
+  - name: "Set default channel when the current_channel is empty"
+    set_fact:
+      current_channel: "{{ (current_channel |length > 0) |
+                            ternary(current_channel,
+                                    package_vars['defaultChannel'] |
```
dirgim, line 29: We also need to check if the defaultChannel isn't an empty string. In case it is, we need to default to the first channel value.

samvarankashyap:
@dirgim PTAL
Fixed it

PR #22, opened 2020-05-26
allow null backport value

scoheb:


roles/parse_operator_bundle/tasks/main.yml
```diff
@@ -20,7 +20,7 @@
     default_channel: "{{ skopeo_inspect_json.Labels['operators.operatorframework.io.bundle.channel.default.v1'] }}"
     current_channel: "{{ skopeo_inspect_json.Labels['operators.operatorframework.io.bundle.channel.default.v1'] }}" # using default channel since the bundle only has one
     ocp_versions: "{{ skopeo_inspect_json.Labels['com.redhat.openshift.versions'].split(',') }}"
-    is_backport: false
+    is_backport: ""
```
lslebodn, line 23: Does empty string mean "not set" ?
If yes would be good to add comment.
Otherwise LGTM

scoheb: added comment

PR #26, opened 2020-06-05
Use first channel from the bundle image channels list as current_channel

dirgim:
* Parse the channels label from bundle image, use first value as current_channel
* Align the parse_operator_bundle role with parse_operator_metadata improvements

### DIFF
--- a/roles/parse_operator_bundle/tasks/main.yml
+++ b/roles/parse_operator_bundle/tasks/main.yml
@@ -1,104 +1,140 @@
 ---
-- name: "Inspect the bundle image with skopeo"
-  shell: "skopeo inspect docker://{{ bundle_image }}"
-  register: skopeo_inspect_result
+- name: "Parsing the operator bundle data"
+  block:
+  - name: "Inspect the bundle image with skopeo"
+    shell: "skopeo inspect docker://{{ bundle_image }}"
+    register: skopeo_inspect_result
 
-- name: "Save the skopeo inspect output to a log file"
-  copy:
-    content: "{{ skopeo_inspect_result.stdout }}"
-    dest: "{{ work_dir }}/bundle-skopeo-inspect.json"
-  when:
-    - skopeo_inspect_result is defined
-    - skopeo_inspect_result.stdout != ""
+  - name: "Save the skopeo inspect output to a log file"
+    copy:
+      content: "{{ skopeo_inspect_result.stdout }}"
+      dest: "{{ work_dir }}/bundle-skopeo-inspect.json"
+    when:
+      - skopeo_inspect_result is defined
+      - skopeo_inspect_result.stdout != ""
 
-- set_fact:
-    skopeo_inspect_json: "{{ skopeo_inspect_result.stdout | from_json }}"
+  - name: "Include the skopeo inspect result vars"
+    set_fact:
+      skopeo_inspect_json: "{{ skopeo_inspect_result.stdout | from_json }}"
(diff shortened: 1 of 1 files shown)


roles/parse_operator_bundle/tasks/main.yml
```diff
@@ -1,104 +1,191 @@
...
+    set_fact:
+      crd_paths: "{{ crd_paths + [item.path] }}"
+    with_items: "{{ crd_paths_result['files'] }}"
+
+  - name: "Include the CSV path into the ansible vars"
+    include_vars:
+      file: "{{ csv_path }}"
+      name: operator_vars
+
+  - name: "Determine and setfact for podname"
```
lslebodn, line 128: podname or  "pod name" ??

dirgim: Missed that one, fixed

roles/parse_operator_bundle/tasks/main.yml
```diff
@@ -1,104 +1,137 @@
...
+  - set_fact:
+      is_bundle_image: "{{ skopeo_inspect_json.Labels['com.redhat.delivery.operator.bundle'] }}"
+      package_name: "{{ skopeo_inspect_json.Labels['operators.operatorframework.io.bundle.package.v1'] }}"
+      default_channel: "{{ skopeo_inspect_json.Labels['operators.operatorframework.io.bundle.channel.default.v1'] }}"
+      channels: "{{ skopeo_inspect_json.Labels['operators.operatorframework.io.bundle.channels.v1'].split(',') }}"
+      ocp_versions: "{{ skopeo_inspect_json.Labels['com.redhat.openshift.versions'].split(',') }}"
+      # if backport label missing, an empty is_backport means not set.
+      is_backport: ""
+
+  - set_fact:
```
samvarankashyap, line 28: @dirgim NIT:  Naming every task would be best practice and standard. 
In this case, there might be a chance the task might fail. 
If that channels variable is empty.
It would be great if we could have a name for this task
like, 
```
name: "fetch the first channel from the channels list"
```.

File: roles/parse_operator_bundle/tasks/main.yml [hunk 3]
```diff
@@ -1,104 +1,137 @@
...
+      current_channel: "{{ channels[0] }}"
+
+  - set_fact:
+      # if backport label missing, an empty is_backport means not set.
```
samvarankashyap, line 32: @dirgim Same as above. It would be great if you could name the task 
with contents in the comment.

roles/parse_operator_bundle/tasks/main.yml
dirgim: Ack, great point - fixed.

roles/parse_operator_bundle/tasks/main.yml
dirgim: Ack

sonam1412:
cvp-redhat-operator-metadata-validation-test failed because `java.lang.IllegalArgumentException: Null value not allowed as an environment variable: CVE_ADVISORY_ID`

sonam1412:
Trigger issue with cvp-isv-operator-metadata-validation-test 
 will be fixed with this MR: https://gitlab.sat.engineering.redhat.com/cvp/pipeline/-/merge_requests/1231

lslebodn:
@dirgim looks like ready to merge.
Or do you wait for something else?

PR #31, opened 2020-06-25
Remove obsolete ubi check

nmars:
Removed the ubi-check files, as they are no longer needed by the image source test.


sonam1412:
> We still use that playbook in sharedLib tests.
> So this change would break it.
> 
> We need to firstly remove it from downstream and then we can consider this playbook as unused.

This function: https://gitlab.sat.engineering.redhat.com/cvp/pipeline/-/blob/master/src/org/cvp/pipeline/OperatorUtils.groovy#L230
is also not being used anywhere in the code except sharedLib. Maybe can be cleaned up from the code base with sharedLib test or a new sharedLib test can be added as per the new modifications been done for the image UBI test

nmars:
ty @sonam1412 & @lslebodn i had forgotten completely about the sharedLib test!

lslebodn:
lgtm

nmars:
The obsolete method in OperatorUtils.groovy and related sharedLib test have been removed from downstream.

PR #36, opened 2020-07-20
Use the oc adm catalog build command in place of offline-cataloger

dirgim:
* Replace offline-cataloger in catalog-initialization test
* Remove offline-cataloger installation from install_operator_prereqs
* Use `oc adm catalog build` command for catalog initialization

### DIFF
--- a/operator-catalog-initialization-test.yml
+++ b/operator-catalog-initialization-test.yml
@@ -8,10 +8,11 @@
     run_upstream: false
     operator_work_dir: "/home/jenkins/agent/test-operator"
     work_dir: "{{ lookup('env', 'WORKSPACE') }}"
-    offline_cataloger_bin_path: "/usr/local/bin/offline-cataloger"
     production_quay_namespace: "redhat-operators"
+    oc_bin_path: "/usr/local/bin/oc"
     go_bin_path: "go"
     check_initializer_result: false
+    run_cleanup: true
 
   tasks:
     - name: "Run operator catalog initialization test on the operator metadata"
--- a/local-test-operator.yml
+++ b/local-test-operator.yml
@@ -43,7 +43,6 @@
         yq_bin_path: "{{ testing_bin_path }}/yq"
         go_bin_path: "{{ testing_bin_path }}/go/bin/go"
         operator_sdk_bin_path: "{{ testing_bin_path }}/operator-sdk"
-        offline_cataloger_bin_path: "offline-cataloger"
 
     - name: "Clear the operator working directory"
       file:
--- a/local.yml
+++ b/local.yml
@@ -30,7 +30,6 @@
     operator_sdk_bin_path: "{{ testing_bin_path }}/operator-sdk"
     umoci_bin_path: "{{ testing_bin_path }}/umoci"
     opm_bin_path: "{{ testing_bin_path }}/opm"
-    offline_cataloger_bin_path: "offline-cataloger"
 
   environment:
     PATH: "{{ ansible_env.PATH }}:{{ ansible_env.HOME }}/.local/bin"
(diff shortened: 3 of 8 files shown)


lslebodn:
I assumr it will fail. Cannot see such option in 3.11 cluster
```
sh-4.2$ oc adm catalog build                        
error: unknown command "catalog build"
See 'oc adm -h' for help and examples.
```

dirgim:
> I assumr it will fail. Cannot see such option in 3.11 cluster
> 
> ```
> sh-4.2$ oc adm catalog build                        
> error: unknown command "catalog build"
> See 'oc adm -h' for help and examples.
> ```

That's true, but it seems like we're able to use the 4.4.x+ `oc` that's available on the operator-utils container to use that command.
If the `--from=scratch` option is set and no `--to` option is set, the `catalog build` command basically works in an `offline` mode and doesn't consult the actual OpenShift cluster, so it works almost exactly the same as `offline-cataloger`.

operator-catalog-initialization-test.yml
```diff
@@ -8,10 +8,11 @@
     run_upstream: false
     operator_work_dir: "/home/jenkins/agent/test-operator"
     work_dir: "{{ lookup('env', 'WORKSPACE') }}"
-    offline_cataloger_bin_path: "/usr/local/bin/offline-cataloger"
     production_quay_namespace: "redhat-operators"
+    oc_bin_path: "/usr/local/bin/oc"
```
lslebodn, line 12: I cannot see any documentation that `oc` has to be stored at this location
Moreover we expect some minimal version of client as well

lslebodn:
Can you share old/new output for random problematic operator?

operator-catalog-initialization-test.yml
dirgim: This playbook is meant for usage in the operator-utils container. In other usages the oc location (and version) will be set as part of the install_prereqs role.

lslebodn: Fair enough.

But then we should document required expectation for local-test-operator.yml

dirgim:
> Can you share old/new output for random problematic operator?

There is no difference in outputs since this change only addresses the downloading of the existing catalog, not the running of the catalog initializer command which tries to add the operator to the catalog. Ex. for 
using this (oc-adm-catalog-init) branch:
> time="2020-08-03T15:50:30+02:00" level=fatal msg="permissive mode disabled" error="error loading manifests from directory: error loading package into db: fuse-console.v7.7.0 specifies replacement that couldn't be found"
using the master branch:
> time="2020-08-03T17:43:01+02:00" level=fatal msg="permissive mode disabled" error="error loading manifests from directory: error loading package into db: fuse-console.v7.7.0 specifies replacement that couldn't be found"

That said, I have also tested and compared the downloaded catalogs for each of the operator types and have found no appreciable differences in the downloaded metadata between implementations (the oc adm method sets the empty status field in CSV files which doesn't affect the catalog initialization - I double checked this with the OLM team).

operator-catalog-initialization-test.yml
dirgim: Added a note to the install_prereqs role.

dirgim:
@samvarankashyap @lslebodn @14rcole PTAL

cvp-ops:
py3 test

14rcole (approved):
LGTM

samvarankashyap (approved):

dirgim:
I have tested the changes on all released operators from all three official catalogs (redhat-operators, certified-operators, community-operators) and all tests were successful. As mentioned above, I also confirmed that the operators that do fail the catalog-initialization test still fail with the same errors. Merging the PR.

PR #41, opened 2020-08-04
Return json instead of error code in parse_operator_metadata

14rcole:
Make the parse_operator_metadata role generate a json file with the results of the playbook and an error message if necessary.  The role also now returns zero regardless of whether there was an error


roles/parse_operator_metadata/tasks/main.yml
```diff
@@ -181,5 +185,13 @@
         msg: "{{ ansible_failed_result }}"
 
     - name: "Result of failure"
-      fail:
-        msg: "Ansible playbook while preparing the operator metadata failed"
+      set_fact:
+        parse_operator_metadata_results: "{{ parse_operator_metadata_results | combine({'result': 'pass', 'msg': ansible_failed_result}) }}"
```
dirgim, line 189: I think `result` should be set to  `failure` here, e.g.
```suggestion
        parse_operator_metadata_results: "{{ parse_operator_metadata_results | combine({'result': 'failure', 'msg': ansible_failed_result}) }}"
```

14rcole: Oversight, thank you

dirgim:
LGTM

sonam1412:
Do we have to update the doc for this change ?

otherwise LGTM

roles/parse_operator_metadata/tasks/main.yml
```diff
@@ -1,3 +1,7 @@
+- name: "Create variables"
+  set_fact:
+    parse_operator_metadata_results: {'result': 'pass', 'msg': ''}
+
```
lslebodn, line 4: It would be better to set `{'result': 'pass', 'msg': ''}` before rescue on line 172
That is the only place we are 100% sure everything is OK.

14rcole:
@sonam1412 Where do I find the upstream doc?

sonam1412:
> @sonam1412 Where do I find the upstream doc?

Thats the readme.md https://github.com/redhat-operator-ecosystem/operator-test-playbooks/blob/master/README.md

14rcole:
@sonam1412 I don't see anything in there that would require updating

sonam1412:
> @sonam1412 I don't see anything in there that would require updating

I donot clearly remember on what we have decided to update in the document (it is mentioned in Jira to update upstream docs) but I think it is to say the o/p will be validated from the result json file instead of the error code.
I request @lslebodn to suggest correct approach

PR #42, opened 2020-08-04
[CVP-1226] operator metadata role update

jsztuka:
This change is primarily made in order to fix following:
https://github.com/operator-framework/community-operators/pull/2014

### DIFF
--- a/roles/inject_scorecard_metadata/tasks/main.yml
+++ b/roles/inject_scorecard_metadata/tasks/main.yml
@@ -48,14 +48,17 @@
   environment:
     PATH: "{{ testing_bin_path}}:{{ lookup('env', 'PATH') }}"
 
-- name: "Load CSV into a csv_vars"
-  include_vars:
-    file: "{{ csv_path }}"
-    name: csv_vars
+- name: "Read csv.yaml file"
+  shell: "cat {{ csv_path }}"
+  register: csv_data
+
+- name: "Set facts for csv_data"
+  set_fact:
+    csv_vars: "{{ csv_data.stdout | from_yaml }}"
 
 - name: "Get spec from csv_vars"
   set_fact:
-    csv_vars_spec: "{{ csv_vars['spec'] | default({}) }}"
+    csv_vars_spec: "{{ (csv_vars | from_yaml)['spec'] | default({}) }}"
 
 - name: "Get customresourcedefinitions from csv_vars_spec"
   set_fact:
--- a/roles/parse_operator_bundle/tasks/main.yml
+++ b/roles/parse_operator_bundle/tasks/main.yml
@@ -89,39 +89,42 @@
       crd_paths: "{{ crd_paths + [item.path] }}"
     with_items: "{{ crd_paths_result['files'] }}"
 
-  - name: "Include the CSV path into the ansible vars"
-    include_vars:
-      file: "{{ csv_path }}"
-      name: operator_vars
+  - name: "Read csv.yaml file"
+    shell: "cat {{ csv_path }}"
+    register: csv_data
+
+  - name: "Set facts for csv_data"
+    set_fact:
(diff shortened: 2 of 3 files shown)


sonam1412:
what's the purpose of log.txt ? maybe got added accidently

jsztuka:
> what's the purpose of log.txt ? maybe got added accidently

Yes, it got involved by accident. Thanks for noticing.

roles/inject_scorecard_metadata/tasks/main.yml
```diff
@@ -48,14 +48,17 @@
...
+  register: csv_data
+
+- name: "Set facts for csv_data"
+  set_fact:
+    csv_vars: "{{ csv_data.stdout | from_yaml }}"
 
 - name: "Get spec from csv_vars"
   set_fact:
-    csv_vars_spec: "{{ csv_vars['spec'] | default({}) }}"
+    csv_vars_spec: "{{ (csv_vars | from_yaml)['spec'] | default({}) }}"
```
sonam1412, line 61: Do we have to convert the csv_vars again here when it is already done on line#57 ?

roles/parse_operator_bundle/tasks/main.yml
```diff
@@ -89,39 +89,42 @@
...
   - name: "Determine and set fact for operator specific information - name, pod name, container name and capabilities"
     set_fact:
-      current_csv: "{{ operator_vars.metadata.name }}"
-      operator_pod_name: "{{ operator_vars.spec.install.spec.deployments[0].name }}"
-      operator_container_name: "{{ operator_vars.spec.install.spec.deployments[0].spec.template.spec.containers[0].name }}"
-      operator_capabilities: "{{ operator_vars.metadata.annotations.capabilities }}"
+      current_csv: "{{ (csv_vars | from_yaml).metadata.name }}"
+      operator_pod_name: "{{ (csv_vars | from_yaml).spec.install.spec.deployments[0].name }}"
+      operator_container_name: "{{ (csv_vars | from_yaml).spec.install.spec.deployments[0].spec.template.spec.containers[0].name }}"
+      operator_capabilities: "{{ (csv_vars | from_yaml).metadata.annotations.capabilities }}"
```
sonam1412, line 105: Do we have to convert the csv_vars again here when it is already done on line#98 ?

roles/parse_operator_metadata/tasks/main.yml
```diff
@@ -118,39 +118,42 @@
...
-    include_vars:
-      file: "{{ csv_path }}"
-      name: operator_vars
+  - name: "Read csv.yaml file"
+    shell: "cat {{ csv_path }}"
+    register: csv_data
+
+  - name: "Set facts for csv_data"
+    set_fact:
+      csv_vars: "{{ csv_data.stdout }}"
```
sonam1412, line 127: Shouldn't this be converted here and use csv_vars at all other recursive places ?

roles/inject_scorecard_metadata/tasks/main.yml
jsztuka: If we used from_yaml only in the initial part of declaring csv_vars and no further, nested variables of that fact(csv_vars) would be unaccessible by ansible.

roles/parse_operator_bundle/tasks/main.yml
jsztuka: Same case as mentioned above.

roles/parse_operator_metadata/tasks/main.yml
jsztuka: Same case as mentioned above.

sonam1412: from_yaml is not needed here?

jsztuka: It is, updated.

sonam1412:
LGTM

PR #48, opened 2020-08-31
parse_operator_metadata - recurse to find pkg path

nmars:
When running `parse-operator-metadata.yml`, the `operator_work_dir` extra-var is specified to indicate the directory where metadata files reside. The `find` module needs to recurse so as to discover the package file if the metadata is contained a sub directory of `operator_work_dir`.


nmars:
I am going to close this. Operator courier expects the package file to reside at the top level directory and not within a sub directory. So the playbook should fail in the same manner.

PR #49, opened 2020-09-15
Bugfix: Fix the false postive runs for parse-operator-metadata role

samvarankashyap:
$subject
This PR fixes the false postive runs for all the metadata-parsing.


dirgim (approved):
LGTM

PR #105, opened 2020-11-25
Add optional-operator-subscribe role into operator-test-playbooks

samvarankashyap:
Add optional-operator-subscribe role in operator-test-playbooks


roles/optional-operators-subscribe/templates/optional-operators-subscribe-commands.sh
```diff
@@ -0,0 +1,176 @@
+#!/bin/bash
```
lslebodn, line 1: I assume this file is almost the same as in project.
Can we add a comment with reference to the source?
Just in case sb would like to keep them in sync.

dirgim (approved):
LGTM

PR #117, opened 2021-01-20
Collect the names of images that were pulled during the operator deployment tests

14rcole:
We currently determine which images were pulled during operator deployment during the operator image source test in CVP, but this approach doesn't work if the tests are run on an external cluster.  These changes save a file with a list of images that were pulled during the deployment test so that the image source test can simply reference the file.

Although the legacy operator deployment test won't be run externally, it still needed to be modified in preparation for the changes that will be made downstream.  This also allows us to implement these changes before the new bundle operator deployment playbooks are complete.

### DIFF
--- a/local-test-operator.yml
+++ b/local-test-operator.yml
@@ -21,6 +21,7 @@
     testing_bin_path: "{{ work_dir }}/bin"
     current_channel: '' # Added to avoid a potential bug with undefined variables
     olm_ver: 0.14.1
+    events_file_path: "{{ work_dir }}/events.json"
 
   environment:
     PATH: "{{ ansible_env.PATH }}:{{ ansible_env.HOME }}/.local/bin"
@@ -150,9 +151,27 @@
             openshift_images_before: " {{ os_registry_is_result_before.stdout_lines }} + {{ os_pod_is_result_before.stdout_lines }}"
           when: not run_upstream|bool
 
+        - name: "Set the deployment start time"
+          shell: "date -u '+%Y-%m-%dT%H:%M:%SZ'"
+          register: start_time
+          when:
+            - run_deploy|bool
+            - not run_upstream|bool
+
         - name: "Deploy the operator on the OpenShift 4.0 with OLM"
           include_role:
             name: deploy_olm_operator
+          when:
+            - run_deploy|bool
+            - not run_upstream|bool
+
+        - name: "Collect the names of the images modified during the deployment"
+          include_role:
+            name: collect_modified_images
+          vars:
+            deployment_start_time: "{{ start_time }}"
+            events_file_path: "{{ events_file_path }}"
+            kubeconfig_path: "{{ kubeconfig_path }}"
           when:
             - run_deploy|bool
             - not run_upstream|bool
--- a/roles/collect_modified_images/tasks/main.yml
+++ b/roles/collect_modified_images/tasks/main.yml
@@ -1,0 +1,37 @@
+---
+# # In order for this role to work, add the following task to your playbook
+# # immediately before calling the deployment role.
+# - name: "Set the deployment start time"
+#   shell: "date -u '+%Y-%m-%dT%H:%M:%SZ'"
+#   register: deployment_start_time
+# 
+# The above shell command gets the date in the same format passed by openshift
+# in the `oc get events` command below.  The below command gets all images
+# pulled to the cluster.  This includes images that were not pulled as part of
+# the deployment.  deployment_start time is used in the parse_image_names file
+# to filter out images pulled before the deployment started.  Since this role
+# should be run immediately after the deployment, there should be no need to filter
+# images pulled after the deployment.
+
+
+
+- name: "Generate events.json"
+  shell: "{{ oc_bin_path }} {% raw %}get events --all-namespaces --field-selector reason==Pulling -o go-template='{{range .items}}{{.lastTimestamp}},{{.message}}{{\"\\n\"}}{{end}}'{% endraw %}"
+  environment:
+    KUBECONFIG: "{{ kubeconfig_path }}"
+  register: events
+
+- name: "get list of images in events.json"
+  include_tasks: parse_image_names.yml
+  with_items:
+    - "{{ events.stdout_lines }}"
+
+- name: "Print images"
+  debug:
+    msg: "pulled images: {{ pulled_images }}"
+
+- name: "Save images to file"
+  module: copy
+  content: "{{ pulled_images | to_json }}"
+  dest: "{{ events_file_path }}"
+  delegate_to: localhost
--- a/roles/collect_modified_images/tasks/parse_image_names.yml
+++ b/roles/collect_modified_images/tasks/parse_image_names.yml
@@ -1,0 +1,24 @@
+# Once the `oc get events` command in tasks/main.yml is run, we will end up
+# with a list of timestamps and pulling messages in CSV format.  main.yml will
+# pass each line into this file for parsing, so "item" will be formatted as
+# follows:
+#
+# 2021-02-15T14:14:19Z,pulling image "docker-registry.default.svc:5000/cvp-ci/cvp-slave:stable"
+#
+# The timestamp is obviously the first item here
+# The second item is the message from which we get the image name.
+# This is hacky, but unfortunately it's the only way to do it right now.
+# The "item.split(',')[1][15:-1]" below gets the substring that contains the
+# image name.  item.split(',')[1] get the message from the item.  [15] is the
+# index of the character after the first quotation mark.  [-1] moves in one
+# character from the right of the string, which chops off the quotation mark at
+# the end of the array.
+- name: "split item"
+  set_fact:
+    timestamp: "{{ item.split(',')[0] }}"
+    image: "{{ item.split(',')[1][15:-1] }}"
+
+- name: "parse and return image name when timestamp is after start_time"
+  set_fact:
+    pulled_images: "{{ pulled_images + [image] }}"
+  when: "{{ timestamp > deployment_start_time }}"
--- a/roles/collect_modified_images/defaults/main.yml
+++ b/roles/collect_modified_images/defaults/main.yml
@@ -1,0 +1,3 @@
+pulled_images: []
+events_file_path: /tmp/events.json
+deployment_start_time: 1970-01-01T00:00:00Z


roles/collect_modified_images/tasks/parse_image_names.yml
```diff
@@ -0,0 +1,9 @@
+- name: "split item"
+  set_fact:
+    timestamp: "{{ item.split(',')[0] }}"
+    image: "{{ item.split(',')[1][15:-1] }}"
+
+- name: "parse and return image name when timestamp is after start_time"
+  set_fact:
+    IMAGES: "{{ IMAGES + [image] }}"
+  when: "{{ timestamp > deployment_start_time }}"
```
samvarankashyap, line 9: @14rcole Does the timestamp match with ansible .iso8601 standard ?? 
Which timezone will be taken in both cases ?
is nt it good for us to use utc times ?

samvarankashyap (commented):
Excecpt for the comment 
PR looks good to me

roles/collect_modified_images/tasks/parse_image_names.yml
14rcole: Both strings are in iso8601, but the timezones are off, good catch.  It looks like openshift uses UTC, but ansible uses the local timezone.  Do you know which version of ansible CVP uses?  There's an easy fix, but it requires 2.8 or higher.

samvarankashyap: @14rcole We are using 2.7 in CVP right now. 
However, In order to get the UTC time i think i have a workaround without moving to 2.8 . 

```yaml
 - name: "Get shell date utc"
      shell: "date -u '+%Y-%m-%d %H:%M:%S'"
      register: date_via_shell
```
You can always get the utc time using shell date command.

collect-modified-images.yml
```diff
@@ -0,0 +1,20 @@
+---
+- name: Collect the names of all images
+  hosts: localhost
+  become: false
+  gather_facts: true
+
+  vars:
+  environment:
+    PATH: "{{ ansible_env.PATH }}:{{ ansible_env.HOME }}/.local/bin"
+      # when: run_upstream|bool
```
nmars, line 10: Should this be un-commented or deleted?

collect-modified-images.yml
```diff
@@ -0,0 +1,20 @@
...
+
+  tasks:
+    - set_fact:
+        # oc_bin_path: '{{ ''kubectl'' if run_upstream else "{{ testing_bin_path }}/oc" }}'
```
nmars, line 14: Lines 14-16 look like they are for testing?

roles/collect_modified_images/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+- name: "Create empty list for images"
+  set_fact:
+    pulled_images: []
+
+- name: "Generate events.json"
+  shell: "{% raw %}oc get events --field-selector reason==Pulling -o go-template='{{range .items}}{{.lastTimestamp}},{{.message}}{{\"\\n\"}}{{end}}' --sort-by={.lastTimestamp}{% endraw %}"
```
nmars, line 6: Should `oc` be hardcoded here, or should we use `oc_bin_path`?

roles/collect_modified_images/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  register: events
+
+- name: "get list of images in events.json"
+  include_tasks: parse_image_names.yml
+  with_items:
+    - "{{ events.stdout_lines }}"
+
+- name: "Print images"
+  debug: 
+  msg: "pulled images: {{ pulled_images }}"
```
nmars, line 16: Need indentation before `msg:`

deploy-olm-operator.yml
```diff
@@ -48,6 +49,10 @@
         name: inject_scorecard_metadata
       when: run_scorecard|bool
 
+    - name: "Set the deployment start time"
+      set_fact:
+        start_time: "{{ ansible_date_time.iso8601 }}"
+
```
lslebodn, line 55: @dirgim IIUC we will very likely use different role for deploying OLM operator for operator bundles.
Is my assumption correct?

roles/collect_modified_images/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- name: "Save images to file"
+  local_action:
+    module: copy
+    content: "{{ pulled_images | join('\n') }}"
```
lslebodn, line 21: I would prefer some structured text as an output instead of single string which has to be parsed

deploy-olm-operator.yml
dirgim: @lslebodn that's correct, this deployment role will be used exclusively for the legacy operator metadata format.
That said, I see no harm of collecting the pulled images as part of this role to provide the legacy pipeline with the same functionality as the new one will have.

lslebodn: And I see no benefit of collecting them. But I am looking forward for reasonable justification of this change.

14rcole: @lslebodn By making this change, we keep all of the code for the image source test downstream the same. I think there's value in that

Of less importance, if we don't implement this here, the ticket associated with this will be blocked until the new playbook is created.

roles/collect_modified_images/tasks/main.yml
14rcole: Each image is listed on a new line, I thought that would be the most straightforward to parse.  If not, I can use some other delineation for the images such as a comma

collect-modified-images.yml
14rcole: This file should actually be deleted come to think of of it

roles/collect_modified_images/tasks/main.yml
14rcole: Nope, good catch.  I changed that for testing

roles/collect_modified_images/tasks/main.yml
lslebodn: It is much reliable to delegate parsing to standard parsers for structural data.

deploy-olm-operator.yml
lslebodn: > @lslebodn By making this change, we keep all of the code for the image source test downstream the same. I think there's value in that

I do not follow here.

It will be just a dead code and increase test-time without any value.

lslebodn: >  if we don't implement this here, the ticket associated with this will be blocked until the new playbook is created.

Ticket is blocked anyway because the scorecard test for operator bundles does not exist yet.
Creating role is useful and should be merged but we will not be able to resolve issue anyway.

roles/collect_modified_images/tasks/main.yml
14rcole: Alright, I can write it to a json file

local-test-operator.yml
```diff
@@ -150,13 +151,24 @@
...
           include_role:
             name: deploy_olm_operator
           when:
             - run_deploy|bool
             - not run_upstream|bool
 
+        - name: "Collect the names of the images modified during the deployment"
+          import_role:
+            name: collect_modified_images
+
```
nmars, line 171: Should this be `include_role` instead of `import_role`?

Should this task also have the same `when` conditional as "Set the deployment start time" above since it depends on `start_time`? 

For clarity I suggest passing `start_time` explicitly by adding
```
vars:
  deployment_start_time: "{{ start_time }}"
```

roles/collect_modified_images/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+# In order for this role to work, add the following task to your playbook
+# immediately before calling the deployment role.
+# - name: "Set the deployment start time"$a
+#   shell: "date -u '+%Y-%m-%dT%H:%M:%SZ'"
+#   register: start_time
+- name: "Create empty list for images"
+  set_fact:
```
nmars, line 8: I think the definition of `pulled_images` and `events_file_path` should be moved to `roles/collect_modified_images/defaults/main.yml` since that file is currently empty and should be used for that purpose.

roles/collect_modified_images/tasks/parse_image_names.yml
```diff
@@ -0,0 +1,9 @@
+- name: "split item"
+  set_fact:
+    timestamp: "{{ item.split(',')[0] }}"
+    image: "{{ item.split(',')[1][15:-1] }}"
+
+- name: "parse and return image name when timestamp is after start_time"
+  set_fact:
+    pulled_images: "{{ pulled_images + [image] }}"
+  when: "{{ timestamp > deployment_start_time }}"
```
nmars, line 9: I couldn't see where `deployment_start_time` is initialized or passed previous to importing these tasks.

deploy-olm-operator.yml
```diff
@@ -2,7 +2,7 @@
 - name: Deploy the operator on an OCP cluster with OLM as part of the operator testing pipeline
   hosts: all
   become: false
-  gather_facts: false
+  gather_facts: true
```
nmars, line 5: We don't need to enable `gather_facts` since this is being run explicitly on localhost and we're not using any ansible* facts (or I think you were but since removed them).

roles/collect_modified_images/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+    pulled_images: []
+    events_file_path: "/tmp/events.txt"
+
+- name: "Generate events.json"
+  shell: "{{ oc_bin_path }} {% raw %}get events --field-selector reason==Pulling -o go-template='{{range .items}}{{.lastTimestamp}},{{.message}}{{\"\\n\"}}{{end}}' --sort-by={.lastTimestamp}{% endraw %}"
```
dirgim, line 13: I believe this command should be run with the `--all-namespaces` option to cover all image pulls on the cluster - operators can manage other namespaces than the one they were installed in.

roles/collect_modified_images/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  register: events
```
dirgim, line 14: This task (Generate events.json) needs to also set the KUBECONFIG variable in its environment to ensure that the correct cluster configuration is being used when running the command. Examples of this can be found in other roles, such as https://github.com/redhat-operator-ecosystem/operator-test-playbooks/blob/master/roles/deploy_olm_operator/tasks/main.yml#L39-L40

local-test-operator.yml
14rcole: I had to import the role instead of including it so that the variables defined within the playbook would be available to the role - I guess passing manually is a bit cleaner

deploy-olm-operator.yml
14rcole: Yes, thank you

lslebodn:
Going to test fedora-messaging changes

roles/collect_modified_images/tasks/parse_image_names.yml
14rcole: It's in local-test-operator.yml when the role is called.  I also added it to `roles/collect_modified_images/defaults.main.yml` just in case

roles/collect_modified_images/tasks/parse_image_names.yml
```diff
@@ -0,0 +1,9 @@
+- name: "split item"
+  set_fact:
+    timestamp: "{{ item.split(',')[0] }}"
+    image: "{{ item.split(',')[1][15:-1] }}"
```
lslebodn, line 4: Such parsing is very dangerous. Input can change due to various reasons. And we might get unexpected results
due to parsing without any validation.

local-test-operator.yml
```diff
@@ -150,13 +151,31 @@
             openshift_images_before: " {{ os_registry_is_result_before.stdout_lines }} + {{ os_pod_is_result_before.stdout_lines }}"
           when: not run_upstream|bool
 
+        - name: "Set the deployment start time"
+          shell: "date -u '+%Y-%m-%dT%H:%M:%SZ'"
```
lslebodn, line 155: Could you explain why we cannot use epoch here?

roles/collect_modified_images/tasks/parse_image_names.yml
14rcole: We're collecting the results as a json file [here](https://github.com/redhat-operator-ecosystem/operator-test-playbooks/blob/4fba992ef95962781bf826da455d8275b1f65500/roles/collect_modified_images/tasks/main.yml#L8) and simply outputting the fields as csv:

```
timestamp1,message1
timestamp2,message2
...
```

The format of the timestamp or message could change, but I don't see a way around that if we're going to parse the events.json.

local-test-operator.yml
14rcole: Do you mean the "seconds since the epoch" format or the literal epoch time?

In the first case, the `oc get events` command I linked in my comment above uses that format for the date.  Getting the date in the same format makes it easier to compare.  The other option is to convert the date from the `oc get events` command to seconds since the epoch.  That requires extra steps 

In the second case, that would give us all images pulled by all namespaces going back as far as the cluster keeps events logs.  The operator bundles will be running on a "clean" cluster, but there could still be images pulled for the cluster deployment that we don't want to drag into the test

lslebodn: Such explanation should be in the code as a comment.
Just in case something will change in the future so we will know why and what need to be changed.

roles/collect_modified_images/tasks/parse_image_names.yml
lslebodn: that is explanation for the 1st index in the split.
but does not explain why do we need `[15:-1]`.

Please put examples as comments into the code.
So sb else will be able to change/fix it if we need. Adding checks for expected format will be even better.

local-test-operator.yml
lslebodn: I still cannot see any explanation in the code.
@14rcole could you explain why did you resolved discussion?

lslebodn: ehm, it was explained in the roles/collect_modified_images/tasks/main.yml

dirgim (approved):
LGTM

nmars:
LGTM

nmars (approved):

PR #119, opened 2021-02-02
[CVP-1610] Fix Operator metadata parsing role which doesn't account for quotes in CSV name

samvarankashyap:
The parse-operator-metadata role fails when CSV name is in quotes. 
This PR fixes it. 
example 
```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: ClusterServiceVersion
metadata:
  name: 'xxxv.6.0'
```
Try, 
```
ansible-playbook -vvvv -i "localhost," --connection=local parse-operator-metadata.yml -e "operator_work_dir=<path to metadata dir>" -e"work_dir=/tmp/test_op/" # any temp dir
```
Actual output of parse_operator_metadata_results.json
```
{
    "msg": {
        "_ansible_no_log": false,
        "changed": false,
        "failed": true,
        "msg": "Unable to find the current csv path please check the csv names in metadata"
    },
    "result": "fail"
}
```

expected output: 
```
{
    "msg": "",
    "result": "pass"
}
```

### DIFF
--- a/roles/parse_operator_metadata/tasks/main.yml
+++ b/roles/parse_operator_metadata/tasks/main.yml
@@ -104,7 +104,7 @@
     register: kind_custerserviceversion_file_paths
 
   - name: "Grep paths with current_csv ClusterServiceVersion"
-    shell: "grep -l 'name: {{ current_csv }}' {{ item.path }}"
+    shell: "grep -lE \"name: '{{ current_csv }}'|name: {{ current_csv }}\" {{ item.path }}"
     with_items: "{{ kind_custerserviceversion_file_paths['files'] }}"
     register: grep_output_current_csv
     ignore_errors: true


samvarankashyap:
@dirgim PTAL

14rcole:
LGTM

14rcole (approved):

roles/parse_operator_metadata/tasks/main.yml
```diff
@@ -104,7 +104,7 @@
     register: kind_custerserviceversion_file_paths
 
   - name: "Grep paths with current_csv ClusterServiceVersion"
-    shell: "grep -l 'name: {{ current_csv }}' {{ item.path }}"
+    shell: "grep -lE \"name: '{{ current_csv }}'|name: {{ current_csv }}\" {{ item.path }}"
```
sonam1412, line 107: I think the second parameter "name:" should have one space for better readability

samvarankashyap: @sonam1412 
According to grep syntax, the pattern changes if i introduce space and meaning changes. 
I tried it. It does not work as expected.

sonam1412: Sure

PR #123, opened 2021-02-16
Ignore CR deletion timeout

nmars:
Made a simple change to ignore errors caused by a hung CR deletion after the scorecard test. The error is not related to the test and allows the playbook to continue and gather all the logs intact.

### DIFF
--- a/roles/operator_scorecard_tests/tasks/main.yml
+++ b/roles/operator_scorecard_tests/tasks/main.yml
@@ -31,10 +31,12 @@
 - debug:
     msg: "Scorecard Basic Tests passed: {{ scorecard_results_passed_result.stdout }}"
 
+  # The scorecard test may have already deleted the CR, so ignore any errors
 - name: "Delete the operator CR"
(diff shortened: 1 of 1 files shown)


dirgim (approved):
LGTM

roles/operator_scorecard_tests/tasks/main.yml
```diff
@@ -35,6 +35,7 @@
   shell: "{{ oc_bin_path }} delete -f {{ cr_path }} --ignore-not-found=true --grace-period=60 --timeout=120s"
   environment:
     KUBECONFIG: "{{ kubeconfig_path }}"
+  ignore_errors: true
```
lslebodn, line 38: Please add the comment why do we ignore errors here.
Neither code nor commit message explain the reason. I was able to find reason in different discussion

nmars: @lslebodn Comment added

lslebodn:
LGTM

PR #124, opened 2021-02-19
[CVP-1644] - get replicaset logs for operators that fail olm-deployment

jsztuka:
Adding a task that saves the output of the command: 
`oc get replicasets -o go-template='{{range .items}}{{.status.conditions}}{{"\n"}}{{end}}' -n $NAMESPACE` 
(adjusted for usage in Ansible) to the deploy-olm-operator role to create more explicit way of debugging issues that occur when olm-deployment test fails and pod doesn't get created.

### DIFF
--- a/roles/deploy_olm_operator/tasks/main.yml
+++ b/roles/deploy_olm_operator/tasks/main.yml
@@ -207,6 +207,19 @@
       set_fact:
         deploy_olm_operator_results: {'result': 'pass', 'msg': ''}
 
+    - name: "Get replicaset logs for operators that failed olm-deployment output"
+      shell: "{{ oc_bin_path }} get replicasets -o go-template={%raw%}'{{range .items}}{{.status.conditions}}{{\"\\n\"}}{{end}}'{%endraw%} -n {{ openshift_namespace }}"
+      register: olm_replicaset_output
+      environment:
+        KUBECONFIG: "{{ kubeconfig_path }}"
+      when: operator_result.rc != 0
+      
+    - name: "Output logs for operators that failed olm-deployment output to a debug file"
+      copy:
+        content: "{{ olm_replicaset_output.stdout }}"
+        dest: "{{ work_dir }}/operator-olm-replicaset-output.txt"
+      when: olm_replicaset_output.stdout is defined
+
     - name: "Failed operator deployment while waiting for operator pod deployment"
       fail:
         msg: 'Operator deployment with OLM failed, check the olm-*.txt files for more details'


roles/deploy_olm_operator/tasks/main.yml
```diff
@@ -233,6 +233,19 @@
     - name: "Failed on deploy_olm_operator_results"
       fail:
         msg: "{{ deploy_olm_operator_results }}"
+    
+    - name: "Get replicaset logs for operators that failed olm-deployment output"
+      shell: "{{ oc_bin_path }} get replicasets -o go-template={%raw%}'{{range .items}}{{.status.conditions}}{{"\n"}}{{end}}'{%endraw%} -n {{ openshift_namespace }}"
+      register: olm_deployment_output
```
dirgim, line 239: Renaming this to `olm_replicaset_output` would make it more clear what was being collected.

roles/deploy_olm_operator/tasks/main.yml
```diff
@@ -233,6 +233,19 @@
...
+      environment:
+        KUBECONFIG: "{{ kubeconfig_path }}"
+      when: deploy_olm_operator_results.result == 'fail'
+    
+    - name: "Output logs for operators that failed olm-deployment output to a debug file"
+      copy:
+        content: "{{ olm_deployment_output.stdout }}"
+        dest: "{{ work_dir }}/operator-olm-deployment-output.txt"
```
dirgim, line 247: Like above, renaming the resulting file to `operator-olm-replicaset-output.txt` would make more sense.

dirgim (changes requested):

roles/deploy_olm_operator/tasks/main.yml
jsztuka: Ok, this filename was mentioned in deliverables in Jira task, changing it.

roles/deploy_olm_operator/tasks/main.yml
jsztuka: I agree.

jsztuka:
Seems like there is currently error with a server
https://jenkins-cvp-stage-ci-operatortestplaybooks.cloud.paas.psi.redhat.com/job/cvp-redhat-operator-bundle-image-validation-test/11/console
Getting 500 - server error.

dirgim (approved):
LGTM

roles/deploy_olm_operator/tasks/main.yml
```diff
@@ -207,6 +207,19 @@
       set_fact:
         deploy_olm_operator_results: {'result': 'pass', 'msg': ''}
 
+    - name: "Get replicaset logs for operators that failed olm-deployment output"
+      shell: "{{ oc_bin_path }} get replicasets -o go-template={%raw%}'{{range .items}}{{.status.conditions}}{{\"\\n\"}}{{end}}'{%endraw%} -n {{ openshift_namespace }}"
```
14rcole, line 211: Don't some operators deploy to multiple namespaces?  Should we be using --all-namespaces?

dirgim: Operators can target and manage multiple namespaces, but are installed in a single namespace - the `openshift_namespace` in our case.

14rcole: Got it

14rcole:
LGTM

14rcole (approved):

PR #129, opened 2021-03-15
[CVP-1683] Failed when CSV.metadata.annotations['alm-examples'] is set to empty

samvarankashyap:
Bugfix: 
Currently, when an ISV operator has the `metadata.annotations.alm-examples` CSV field as empty, the operator-sdk scorecard (which expects the field to contain CR examples in JSON form) fails when trying to run.
the following PR fixes the above issue.


14rcole (approved):
LGTM

dirgim (approved):
LGTM

PR #130, opened 2021-03-16
[CVP-1683]Add tests to empty annotations in metadata

samvarankashyap:
$subject 
Still work in progress 
helps to test existing GitHub actions too.

### DIFF
--- a/.github/workflows/build_and_test_cvp_image.yml
+++ b/.github/workflows/build_and_test_cvp_image.yml
@@ -1,0 +1,59 @@
+name: "Build and test CVP CI image"
+
+# Controls when the action will run. 
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  pull_request:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      # Runs a set of commands using the runners shell
+      - name: "Build the cvp_image inside Dockerfiles"
+        run: |
+          podman --version
+          cd Dockerfiles 
+          podman build -t cvp_image:latest -f Dockerfile.cvp --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=master
+
+      - name: "Run example bundle through cvp_image -> POSITVE test"
+        run: |
+          cd Dockerfiles
+          mkdir output_logs
+          podman run -it -v ./example-bundle:/project/operator-bundle -v ./output_logs:/project/output --security-opt label=disable cvp_image:latest
+          # clean output logs for next step
+          rm -rf output_logs
+
+      - name: "Run example bundle through cvp_image --> negative test"
+        continue-on-error: true
+        run: |
+          # cd into Dockerfiles since it resets everystep
+          cd Dockerfiles
+          mkdir empty-example-bundle output_logs
+          podman run -it -v ./empty-example-bundle:/project/operator-bundle -v ./output_logs:/project/output --security-opt label=disable cvp_image:latest
+
+      - name: "Fail when previous step does not fail"
+        run: |
+          grep -inr "metadata not found in ../operator-bundle" Dockerfiles/output_logs/
+
+      - name: "Run example operator-metadata through cvp_image --> negative test"
+        continue-on-error: true
+        run: |
+          # cd into Dockerfiles since it resets everystep
+          cd Dockerfiles
+          mkdir test_operator_work_dir
+          podman run -it -v ./example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=prepare_operator_metadata --security-opt label=disable cvp_image:latest
+
+      - name: "Fail when previous step does not fail"
+        run: |
+          grep -inr "Failed due to CSV.metadata.annotations\['alm-examples'\] is set to empty" Dockerfiles/output_logs
--- a/Dockerfiles/Dockerfile.cvp
+++ b/Dockerfiles/Dockerfile.cvp
@@ -1,0 +1,16 @@
+FROM registry.fedoraproject.org/fedora:33
+WORKDIR /project
+ARG OPERATOR_SDK_VERSION=v1.4.0
+ARG OPERATOR_TEST_PLAYBOOKS_TAG=v1.0.11
+RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(arch) ;; esac);\
+    export OS=$(uname | awk '{print tolower($0)}');\
+    export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/;\
+    curl -L -o /usr/local/bin/operator-sdk ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH} \ 
+    && chmod a+x /usr/local/bin/operator-sdk && \
+    mkdir /project/output && \
+    dnf install --setopt=install_weak_deps=False -y git-core ansible python3-pip mlocate && \
+    dnf clean all && \
+    pip3 install operator-courier && \
+    git clone --branch $OPERATOR_TEST_PLAYBOOKS_TAG https://github.com/redhat-operator-ecosystem/operator-test-playbooks.git
+COPY ./entrypoint_cvp.sh /project/entrypoint_cvp.sh
+ENTRYPOINT ["/project/entrypoint_cvp.sh"]
--- a/Dockerfiles/entrypoint_cvp.sh
+++ b/Dockerfiles/entrypoint_cvp.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+# operator-bundle-image validate
+validate_operator_bundle () {
+    echo -e "operatorSDKVersion is $operatorSDKVersion"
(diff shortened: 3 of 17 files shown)


lslebodn:
You need to fix the playbook as well

The "empty annotation" change broke contact in downstream.
```
04:50:28  ERROR: No artifacts found that match the file pattern "operators/config/ansible/parse_operator_metadata_results.json". Configuration error?
04:50:28  Finished: FAILURE

```

Dockerfiles/Dockerfile
```diff
@@ -8,8 +8,9 @@ RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm6
     curl -L -o /usr/local/bin/operator-sdk ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH} \ 
     && chmod a+x /usr/local/bin/operator-sdk && \
     mkdir /project/output && \
-    dnf install --setopt=install_weak_deps=False -y git-core ansible && \
+    dnf install --setopt=install_weak_deps=False -y git-core ansible python3-pip mlocate && \
     dnf clean all && \
+    pip3 install operator-courier && \
     git clone --branch $OPERATOR_TEST_PLAYBOOKS_TAG https://github.com/redhat-operator-ecosystem/operator-test-playbooks.git
 COPY ./entrypoint.sh /project/entrypoint.sh
 ENTRYPOINT ["/project/entrypoint.sh"]
```
lslebodn, line 16: Why do we change this file ?
operator-courier is unrelated to emty annotation change.

Dockerfiles/entrypoint.sh
```diff
@@ -1,23 +1,42 @@
...
+elif [[ "$TEST_NAME_ENV" = 'prepare_operator_metadata' ]]; then
+    echo "TEST_NAME is set to $TEST_NAME_ENV"
+    echo "Running prepare_operator_metadata"
+    prepare_operator_metadata
 else
-  echo -e "\nBundle image validation Failed"
-  exit 1
+    echo "TEST_NAME is set to $TEST_NAME_ENV"
+    echo "$TEST_NAME_ENV not found in entrypoint.sh"
 fi
```
lslebodn, line 42: Why do we change this file ?
It should be used just for midstream and it was not planned for the role roles/parse_operator_metadata

Dockerfiles/Dockerfile
samvarankashyap: In order to test an empty annotation file, it has to go through prepare-operator-metadata role. 
Which requires operator-courier
https://github.com/redhat-operator-ecosystem/operator-test-playbooks/blob/master/prepare-operator-metadata.yml
Which needs operator-courier to nest copy the metadata. 
I tested the role with the same playbook thus requiring prepare-operator-metadata.yml

Dockerfiles/entrypoint.sh
samvarankashyap: I am making use of the same image without changing its functionality for testing. 
Do you think we need a separate image for testing? 
By default github actions run in ubuntu. 
Inorder to test it inside fedora am using podman to run inside fedora 33 container. 
Am using the same image to test our playbooks. 
Since the build time allocated to opensource projects is 1000 minutes for github actions afaik 
I am keeping it as minimum as possible.

samvarankashyap:
> 04:50:28  ERROR: No artifacts found that match the file pattern "operators/config/ansible/parse_operator_metadata_results.json". Configuration error?

@lslebodn Could you share the downstream link with me? 
As far as I see the all downstream tests went fine when I created the pull request.

Dockerfiles/entrypoint.sh
lslebodn: > Do you think we need a separate image for testing?

yes; midstream image should be used just for its purpose. Testing in midstream and it should be as small as possible.


> Since the build time allocated to opensource projects is 1000 minutes for github actions afaik
16 minutes is enough for unit-tests (including building the image)
Unit test have to be fast.

Dockerfiles/Dockerfile
lslebodn: The same commend as in different thread. the Dockerfle for midstream must be used just for midstream and nothing else.

lslebodn:
> @lslebodn Could you share the downstream link with me?
> As far as I see the all downstream tests went fine when I created the pull request.

Check gchat complains from users.

samvarankashyap:
@lslebodn 
I have fixed the bug with this PR PTAL

https://github.com/redhat-operator-ecosystem/operator-test-playbooks/pull/131

.github/workflows/build_and_test_cvp_image.yml
```diff
@@ -0,0 +1,61 @@
+# This is a basic workflow to help you get started with Actions
```
14rcole, line 1: Delete this line

samvarankashyap: done

dirgim (approved):
LGTM

Dockerfiles/Dockerfile.cvp
```diff
@@ -0,0 +1,16 @@
+FROM registry.fedoraproject.org/fedora:33
```
lslebodn, line 1: The name of dockerfile is really confusing. It has nothing to do with CVP.
It should be just about CI. Moreover I do not see an reason why we should mix files for midstream and CI in the same directory. Files for midstream should be just in its own directory.

And files for CI in different one. And ideally name of directory should be self explanatory. The current name "Dockerfiles" does nto say much to me.

Dockerfiles/Dockerfile.cvp
```diff
@@ -0,0 +1,16 @@
...
+    export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/;\
+    curl -L -o /usr/local/bin/operator-sdk ${OPERATOR_SDK_DL_URL}/operator-sdk_${OS}_${ARCH} \ 
+    && chmod a+x /usr/local/bin/operator-sdk && \
+    mkdir /project/output && \
+    dnf install --setopt=install_weak_deps=False -y git-core ansible python3-pip mlocate && \
+    dnf clean all && \
+    pip3 install operator-courier && \
+    git clone --branch $OPERATOR_TEST_PLAYBOOKS_TAG https://github.com/redhat-operator-ecosystem/operator-test-playbooks.git
+COPY ./entrypoint_cvp.sh /project/entrypoint_cvp.sh
+ENTRYPOINT ["/project/entrypoint_cvp.sh"]
```
lslebodn, line 16: Why do we need entrypoint? We can manually specify file when running container

Dockerfiles/entrypoint_cvp.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# operator-bundle-image validate
+validate_operator_bundle () {
+    echo -e "operatorSDKVersion is $operatorSDKVersion"
+    ansible-playbook -vvvv --connection local -i localhost, operator-test-playbooks/validate-operator-bundle.yml -e"work_dir=/project/output" -e"operator_work_dir=/project/operator-bundle/"
```
lslebodn, line 6: The empty  annotation was not added for operator bundles.
why do we touch bundle here?

.github/workflows/build_and_test_cvp_image.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: |
+          podman --version
+          cd Dockerfiles 
+          podman build -t cvp_image:latest -f Dockerfile.cvp --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=master
+
+      - name: "Run example bundle through cvp_image -> POSITVE test"
+        run: |
+          cd Dockerfiles
+          mkdir output_logs
+          podman run -it -v ./example-bundle:/project/operator-bundle -v ./output_logs:/project/output --security-opt label=disable cvp_image:latest
```
lslebodn, line 33: IIRc ubuntu uses apparmor and not SElinux. And disabling of apparmor shoudl be done with `--security-opt "apparmor=unconfined"`

.github/workflows/build_and_test_cvp_image.yml
```diff
@@ -0,0 +1,59 @@
...
+        continue-on-error: true
+        run: |
+          # cd into Dockerfiles since it resets everystep
+          cd Dockerfiles
+          mkdir test_operator_work_dir
+          podman run -it -v ./example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=prepare_operator_metadata --security-opt label=disable cvp_image:latest
+
+      - name: "Fail when previous step does not fail"
+        run: |
+          grep -inr "Failed due to CSV.metadata.annotations\['alm-examples'\] is set to empty" Dockerfiles/output_logs
```
lslebodn, line 59: We need to check whether all expected files were generated. It is not enough to check message in error log.

Moreover current approach for "running" tests does not allow local execution of tests. This is huge disadvantage.
Because you can run "tests" just in github actions.

Dockerfiles/Dockerfile.cvp
samvarankashyap: IMO Entrypoint acts as starting point execution. It can be extended in near future to add multiple tests. 
we can just add a function in a shell script to add multiple tests.

Dockerfiles/entrypoint_cvp.sh
samvarankashyap: I borrowed it from the midstream CI image to include the test. 
I can remove it. It acts as a unit test for ansible-playbooks we have right now. 
I can remove it if you dont want it here.

.github/workflows/build_and_test_cvp_image.yml
samvarankashyap: I will verify this.

.github/workflows/build_and_test_cvp_image.yml
samvarankashyap: Local execution of tests can be avoided because whenever you fork this repository all the tests will be added to the fork. if you push anything to your fork you should be able to run it. 
However, We can run the tests locally by running the commands inside GitHub actions locally. 
There is no resitriction.

Dockerfiles/Dockerfile.cvp
samvarankashyap: Dockerfiles is a convention I followed from other projects on Github. 
We can definitely segregate the files to make them more comprehendible.

PR #132, opened 2021-03-23
Re-organize the folder structure and make upstream tests run on checked out code

samvarankashyap:
The following PR organizes the existing code base for GitHub workflows 
makes sure the ci images stay separately. 
Further, It also fixes the bug where the tests for operator-metadata runs 
on the checked out code of pull request rather than code built on the tag.


.github/workflows/build_and_test_upstream_ci_image.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          mkdir test_operator_work_dir output_logs
+          podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=prepare_operator_metadata upstream_ci_image:latest
+
+      - name: "Fail when previous step does not fail"
+        run: |
+          grep -inr "Failed due to CSV.metadata.annotations\['alm-examples'\] is set to empty" ./output_logs
+          ls output_logs
+          test -f ./output_logs/parse_operator_metadata_results.json
+          test -f ./output_logs/parsed_operator_data.yml
```
lslebodn, line 38: How will I run these assertions locally as part of development?

Dockerfiles/ci/Dockerfile
```diff
@@ -10,7 +9,6 @@ RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm6
     mkdir /project/output && \
     dnf install --setopt=install_weak_deps=False -y git-core ansible python3-pip mlocate && \
     dnf clean all && \
-    pip3 install operator-courier && \
-    git clone --branch $OPERATOR_TEST_PLAYBOOKS_TAG https://github.com/redhat-operator-ecosystem/operator-test-playbooks.git
-COPY ./entrypoint_cvp.sh /project/entrypoint_cvp.sh
-ENTRYPOINT ["/project/entrypoint_cvp.sh"]
+    pip3 install operator-courier
+COPY ./entrypoint.sh /project/entrypoint.sh
+ENTRYPOINT ["/project/entrypoint.sh"]
```
lslebodn, line 14: I am sorry but I still do not think that ENTRYPOINT is ideal here.

Sure you can override it with podman run --entrypoint.
But we will like to run more tests in future.

The current version of "test" expect execution as part of "entrypoint" and assertions are executed later.
that will complicate running more independent tests with the same entrypoint.
And you can have just single entrypoint in dokcerfile

https://docs.docker.com/engine/reference/builder/#entrypoint

https://docs.docker.com/engine/reference/builder/#cmd

.github/workflows/build_and_test_upstream_ci_image.yml
samvarankashyap: These assertions will check for the file's existence and fail if they do not exist.

Dockerfiles/ci/Dockerfile
samvarankashyap: @lslebodn So you would like to have all the assertions also as a part of the container run such that 
the container ran locally would run the tests and assertions together.

Dockerfiles/ci/Dockerfile
```diff
@@ -10,7 +9,7 @@ RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm6
     mkdir /project/output && \
     dnf install --setopt=install_weak_deps=False -y git-core ansible python3-pip mlocate && \
     dnf clean all && \
-    pip3 install operator-courier && \
-    git clone --branch $OPERATOR_TEST_PLAYBOOKS_TAG https://github.com/redhat-operator-ecosystem/operator-test-playbooks.git
-COPY ./entrypoint_cvp.sh /project/entrypoint_cvp.sh
-ENTRYPOINT ["/project/entrypoint_cvp.sh"]
+    pip3 install operator-courier
+ADD ./command.sh /command.sh
+RUN chmod +x /command.sh
```
lslebodn, line 14: Use executable bit it git repo instead of running it as separate RUN directive

lslebodn:
@samvarankashyap looks liek you did not get the main point from the review.
Using ENTRYPOINT, CMD or manually executing command with podman run is just cosmetic change.

The main issue is with testing itself.

ATM, the file Dockerfiles/ci/command.sh covers execution of whole testsuite and then you run "assertions"  in following "stage" in github actions. In another words, you use github actions as testing framework.

Test suite usually consist of many tests. Assertions are used directly inside each tests. Or immediately after execution test (behavior driven development).

Github action is CI/CD platform and not testing framework. We needn't use real frameworks from python, java, ...
We can use bash for that. But "assertions" should not be written in github actions.

samvarankashyap:
@lslebodn PTAL

Dockerfiles/ci/command.sh
```diff
@@ -0,0 +1,40 @@
...
+        exit 0
+    else
+	echo "test_for_alm_examples Test failed"
+        exit 1
+    fi	
+}
+
+
+TEST_NAME_ENV="${TEST_NAME:-NOTEST}"
+if [[ "$TEST_NAME_ENV" = 'test_for_report_failed_empty_alm_examples' ]]; then
```
lslebodn, line 32: just a question. Should we run all tests when TEST_NAME_ENV is not specified?

IMHO it is bit unclear that `test_prepare_operator_metadata` was executed for
`-e TEST_NAME=test_for_report_failed_empty_alm_examples`

samvarankashyap: @lslebodn Since we do not have any other test i have mentioned only 
```
test_for_report_failed_empty_alm_examples
```
I can default it to all tests if not mentioned.

samvarankashyap: I have changed the default to run all the tests 
@lslebodn

lslebodn: Thank you for changing default for missing env.

But Do we really need to run `test_prepare_operator_metadata` for `-e TEST_NAME=test_for_report_failed_empty_alm_examples`

I would expect just single test would be executed if I specify some name in env variable `TEST_NAME_ENV`

lslebodn: At least that is how I understand the use-case from the name of variable,

If my understanding is not clear than you need to either update name of variables
or to better document the use-case

Dockerfiles/ci/command.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
jsztuka, line 1: I'm just curious, what is the purpose of this shell file? Since the name is command.sh, I don't know what the meaning is.

jsztuka: I see it runs playbook, testing alm examples, then test env name, but what is the general meaning?

samvarankashyap: This command.sh acts as the running command in Dockerfile. 
whenever the container is run. This command.sh triggers.

Dockerfiles/ci/command.sh
samvarankashyap: I will document the flow. 
we need to run the test_prepare_operator_metadata since it would create all the output logs. 
I can make it a part of the test_for_report_failed_empty_alm_examples 
That can clear out the confusion. Further, 
I will re-name the  test_prepare_operator_metadata  to prepare_operator_metadata.

samvarankashyap: Done. Added documentation. 
Please re-review. Thanks.

lslebodn: It still does not make a sense to me.

Let say that we have 5 tests:
* test_for_report_failed_empty_alm_examples
* test_2
* test_3
* test_4
* test_5.

I would expect that all tests will be executed for missing env TEST_NAME_ENV
for  `e TEST_NAME_ENV=test_5` jsut the *test_5* will be executed.

With the current logic everything would be executed because `"test_5" != 'test_for_report_failed_empty_alm_examples'`

I am sorry but that does not make any sense to me.
I cannot see any benefit of this `if/else` block with he current implmentation

Dockerfiles/ci/command.sh
```diff
@@ -0,0 +1,43 @@
...
+    test -f $WORK_DIR/parsed_operator_data.yml
+    test -f $WORK_DIR/parse_operator_metadata_results.json
+
+}
+
+
+test_for_report_failed_empty_alm_examples() {
+    echo "test_for_report_failed_empty_alm_examples: This test checks for failure when empty alm examples are found"
+    prepare_operator_metadata
+    grep -inr "Failed due to CSV.metadata.annotations\['alm-examples'\] is set to empty" $WORK_DIR
```
lslebodn, line 21: We most not use recursive grep. 
We already know which file should contain the message. And we have to check that in the expected file and not anywhere.

Dockerfiles/ci/command.sh
```diff
@@ -0,0 +1,43 @@
...
+OPERATOR_WORK_DIR="${OPERATOR_WORK_DIR:-/project/test_operator_work_dir/}"
+WORK_DIR="${WORK_DIR:-/project/output/}"
+
+# prepares operator metadata for a give operator bundle based on 
+# $OPERATOR_DIR environment variable
+prepare_operator_metadata () {
+    ansible-playbook -vvv -i localhost, --connection local operator-test-playbooks/prepare-operator-metadata.yml -e "operator_dir=$OPERATOR_DIR" -e "operator_work_dir=$OPERATOR_WORK_DIR" -e"work_dir=$WORK_DIR"
+    cp /project/operator-test-playbooks/parse_operator_metadata_results.json $WORK_DIR
+    test -f $WORK_DIR/parsed_operator_data.yml
+    test -f $WORK_DIR/parse_operator_metadata_results.json
```
lslebodn, line 13: We should fail if files do not exist.
We wilt just ignore failure with the current implementation

Dockerfiles/ci/command.sh
samvarankashyap: If there are more tests there will be more elif statements inside commands.sh
By default, if there is no test specified, test_name_env is set to NOTEST and it will default to else block 
, and else block will run all tests. 
@lslebodn Did this clarify ? 
Shall i use a switch statement instead? 
how would you like this to be ?

samvarankashyap: If the environment variable is not set 
line 31: TEST_NAME_ENV="${TEST_NAME:-NOTEST}"
 it would be initialized to 'NOTEST' value. 
 @lslebodn

lslebodn: > If there are more tests there will be more elif statements inside commands.sh

Such approach does not scale.
I cannot imagine  very long `if/elif/else`block for 100+ tests

Dockerfiles/ci/command.sh
samvarankashyap: @lslebodn In recent commit, I have made use of the python unit test framework as CMD inside Dockerfile. 
I felt it would be much easier for us to run/write unit tests. 
Please review. 
If approved I can squash both the commits.

Dockerfiles/ci/run_tests.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_for_report_failed_empty_alm_examples(self):
+        EXEC_CMD = "ansible-playbook -vvv -i localhost, --connection local \
+                    operator-test-playbooks/prepare-operator-metadata.yml \
+                    -e 'operator_dir={operator_dir}' \
+                    -e 'operator_work_dir={operator_work_dir}' \
+                    -e'work_dir={work_dir}'".format(operator_dir=self.OPERATOR_DIR,
+                                                    operator_work_dir=self.OPERATOR_WORK_DIR,
+                                                    work_dir=self.WORK_DIR)
+        os.system(EXEC_CMD)
```
lslebodn, line 28: Can we use subprocess.run?
https://docs.python.org/3/library/subprocess.html#module-subprocess

Dockerfiles/ci/run_tests.py
```diff
@@ -0,0 +1,41 @@
...
+                                      "/project/operator_dir/")
+        self.OPERATOR_WORK_DIR = os.getenv('OPERATOR_WORK_DIR',
+                                           "/project/test_operator_work_dir/")
+        self.WORK_DIR = os.getenv('WORK_DIR',
+                                  "/project/output/")
+        self.PLAYBOOKS_DIR = os.getenv('PLAYBOOKS_DIR',
+                                       "/project/operator-test-playbooks/")
+
+    def test_for_report_failed_empty_alm_examples(self):
+        EXEC_CMD = "ansible-playbook -vvv -i localhost, --connection local \
```
lslebodn, line 21: Do not use upper case names for local variables

Dockerfiles/ci/run_tests.py
```diff
@@ -0,0 +1,41 @@
...
+class RunOperatorTestPlaybookTests(unittest.TestCase):
+
+    def setUp(self):
+        self.OPERATOR_DIR = os.getenv('OPERATOR_DIR',
+                                      "/project/operator_dir/")
+        self.OPERATOR_WORK_DIR = os.getenv('OPERATOR_WORK_DIR',
+                                           "/project/test_operator_work_dir/")
+        self.WORK_DIR = os.getenv('WORK_DIR',
+                                  "/project/output/")
+        self.PLAYBOOKS_DIR = os.getenv('PLAYBOOKS_DIR',
```
lslebodn, line 17: Do not use uppercase names for clamm members

run_tests.py:11:8: C0103: Attribute name "OPERATOR_DIR" doesn't conform to snake_case naming style (invalid-name)
run_tests.py:13:8: C0103: Attribute name "OPERATOR_WORK_DIR" doesn't conform to snake_case naming style (invalid-name)
run_tests.py:15:8: C0103: Attribute name "WORK_DIR" doesn't conform to snake_case naming style (invalid-name)
run_tests.py:17:8: C0103: Attribute name "PLAYBOOKS_DIR" doesn't conform to snake_case naming style (invalid-name)

Dockerfiles/ci/run_tests.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+
+import unittest
+import os
+import os.path
+from os import path
```
lslebodn, line 6: Why do we import path twice ?

Dockerfiles/ci/run_tests.py
```diff
@@ -0,0 +1,41 @@
...
+                    -e 'operator_dir={operator_dir}' \
+                    -e 'operator_work_dir={operator_work_dir}' \
+                    -e'work_dir={work_dir}'".format(operator_dir=self.OPERATOR_DIR,
+                                                    operator_work_dir=self.OPERATOR_WORK_DIR,
+                                                    work_dir=self.WORK_DIR)
+        os.system(EXEC_CMD)
+        self.assertEqual(True,
+                         path.exists("{}/parsed_operator_data.yml".format(self.WORK_DIR)))
+        self.assertEqual(True,
+                         path.exists("{}/parse_operator_metadata_results.json".format(self.PLAYBOOKS_DIR)))
```
lslebodn, line 32: There is even `assertTrue`; you needn't use `self.assertEqual(True,`

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue

Dockerfiles/ci/run_tests.py
```diff
@@ -0,0 +1,41 @@
...
+        fd = open("{}/parse_operator_metadata_results.json".format(self.PLAYBOOKS_DIR), "r")
+        result_output = fd.read()
+        fd.close()
+        print(result_output)
```
lslebodn, line 36: We should read json file as json. And please use `with` statement with open
https://www.pythonforbeginners.com/files/with-statement-in-python

```
with open('some.json', "r") as fd:
    data = json.load(fd)
```

Dockerfiles/ci/command.sh
lslebodn: Using python unitest looks much better.

Dockerfiles/ci/run_tests.py
samvarankashyap: done

Dockerfiles/ci/run_tests.py
samvarankashyap: updated

Dockerfiles/ci/run_tests.py
samvarankashyap: updated

Dockerfiles/ci/run_tests.py
samvarankashyap: updated

Dockerfiles/ci/run_tests.py
samvarankashyap: updated

Dockerfiles/ci/run_tests.py
samvarankashyap: updated

samvarankashyap:
@lslebodn PTAL

Dockerfiles/ci/run_tests.py
```diff
@@ -0,0 +1,40 @@
...
+                                                    operator_work_dir=self.operator_work_dir,
+                                                    work_dir=self.work_dir)
+        subprocess.run(exec_cmd, shell=True)
+        self.assertTrue(path.exists("{}/parsed_operator_data.yml".format(self.work_dir)))
+        self.assertTrue(path.exists("{}/parse_operator_metadata_results.json".format(self.playbooks_dir)))
+        with open("{}/parse_operator_metadata_results.json".format(self.playbooks_dir), "r") as fd:
+            result_output = json.loads(fd.read())
+            print(result_output)
+            failed = "Failed due to CSV.metadata.annotations['alm-examples'] is set to empty" in result_output["msg"]["msg"]
+            self.assertTrue(failed)
```
lslebodn, line 37: `assertIn` should give a better message in case of failure.

https://docs.python.org/3/library/unittest.html#unittest.TestCase.debug

Dockerfiles/ci/run_tests.py
```diff
@@ -0,0 +1,40 @@
...
+                    -e 'operator_dir={operator_dir}' \
+                    -e 'operator_work_dir={operator_work_dir}' \
+                    -e'work_dir={work_dir}'".format(operator_dir=self.operator_dir,
+                                                    operator_work_dir=self.operator_work_dir,
+                                                    work_dir=self.work_dir)
+        subprocess.run(exec_cmd, shell=True)
+        self.assertTrue(path.exists("{}/parsed_operator_data.yml".format(self.work_dir)))
+        self.assertTrue(path.exists("{}/parse_operator_metadata_results.json".format(self.playbooks_dir)))
+        with open("{}/parse_operator_metadata_results.json".format(self.playbooks_dir), "r") as fd:
+            result_output = json.loads(fd.read())
```
lslebodn, line 34: json module has tfow functions `json.load` and `json.loads`
the 1st one allow to pass object with read function
https://docs.python.org/3/library/json.html#json.load

`json.loads(fd.read())` -> `json.load(fd)`

lslebodn:
last two nitpicks; otherwise lgtm

.github/workflows/build_and_test_upstream_ci_image.yml
```diff
@@ -0,0 +1,31 @@
+name: "Build and test upstream CI image"
+
+# Controls when the action will run.
```
14rcole, line 3: nit: I'm not sure how much of this 'boilerplate documentation we need.  If anything, I think the format of the github actions workflows should be documented in one place to make the yaml files a bit easier to read.  I'm open to suggestions.  If you think it should be kept here that's fine too.

Dockerfiles/ci/README.md
```diff
@@ -0,0 +1,37 @@
+# CVP upstream image for testing in github actions
+CVP upstream image operator bundle validate container
```
nmars, line 2: Could we make this more descriptive? Like "This readme describes how to build and execute a container that validates an upstream operator bundle."

Dockerfiles/ci/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Usage
```
nmars, line 4: Nit: Is `#Usage` necessary? We already have a first-level heading

Dockerfiles/ci/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### podman build -t <image_name>:<tagname> -f <filepath> --build-arg OPERATOR_SDK_VERSION=<operator_sdk_version>
+### example:
+podman build -t upstream_ci_image:latest -f Dockerfiles/ci/Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0
+```
+
+## Running tests locally with built image
+
+```
+// create empty directories for testing
```
nmars, line 20: Suggest `// Create empty directories to hold artifacts from the test container`

Dockerfiles/ci/README.md
```diff
@@ -0,0 +1,37 @@
...
+mkdir test_operator_work_dir output_logs
+podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_failed_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
```
nmars, line 22: Suggest adding a comment here for consistency, like `// Run the test container. Artifacts will be saved in the mounted directories.`

Also, unless this is covered someplace else, it would be good to describe what each of the environment variables are for. Like, maybe explain the difference between OPERATOR_DIR and OPERATOR_WORK_DIR.

Dockerfiles/ci/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+```
```
nmars, line 24: Does this section need to be within a code fence "```" ?

Dockerfiles/ci/README.md
```diff
@@ -0,0 +1,37 @@
...
+Note: currently there is only one test inside the container image i.e., test_for_report_failed_empty_alm_examples
+Even though the environment variable is not mentioned container tries to run all the tests by default.
```
nmars, line 26: I don't know what environment variable you are referring to. Could you flesh this out a bit?

Dockerfiles/ci/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+```
+// create empty directories for testing
+mkdir test_operator_work_dir output_logs
+podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_failed_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
+```
+```
+Note: currently there is only one test inside the container image i.e., test_for_report_failed_empty_alm_examples
+Even though the environment variable is not mentioned container tries to run all the tests by default. 
+In order to add more tests please add more functions to command.sh inside the Dockerfiles/ci folder in operator-test-playbooks
```
nmars, line 27: I'm guessing you mean the new test python script and not `command.sh`?

.github/workflows/build_and_test_upstream_ci_image.yml
samvarankashyap: Inline documentation always help. 
I love the idea of adding github action workflows documentation. 
I will try to add separate document about github actions not as a part of this PR .

Dockerfiles/ci/README.md
samvarankashyap: Outdated docs will update.

lslebodn (approved):
lgtm

jsztuka (approved):
LGTM

PR #142, opened 2021-04-22
Fix output dir for store the result in a JSON file

p3ck:
In optional operators subscribe the task "Store the result in a json file"
does not specify the ARTIFACT_DIR.  This may fail if the current working
directory is not writable and we can't find it if it's not in the
ARTIFACT_DIR.

### DIFF
--- a/roles/optional-operators-subscribe/tasks/main.yml
+++ b/roles/optional-operators-subscribe/tasks/main.yml
@@ -34,6 +34,6 @@
   always:
   - name: "Store the result in a JSON file"
     copy:
-      dest: "optional_operator_subscribe.json"
+      dest: "{{ ARTIFACT_DIR }}/optional_operator_subscribe.json"
       content: "{{ shell_script_output }}"
       mode: 0644


14rcole (approved):
LGTM

p3ck:
The CI failures don't look related to my change.  How do we go about getting those resolved so this can be merged?

dirgim (approved):
LGTM

dirgim:
@p3ck the fix for CI issues is in #143

roles/optional-operators-subscribe/tasks/main.yml
```diff
@@ -34,6 +34,6 @@
   always:
   - name: "Store the result in a JSON file"
     copy:
-      dest: "optional_operator_subscribe.json"
```
lslebodn, line 37: AMT default value for two directories are the same.
```
      ARTIFACT_DIR: "{{ ARTIFACT_DIR }}"
      SHARED_DIR: "{{ ARTIFACT_DIR }}"
```

Storing json to directory with artifacts make a sense. And all of them should be included in  tarball created by your team. Do we really need to collect also data from "SHARED_DIR"

Should be one sub-directory or other one ? @dirgim You introduced the `"SHARED_DIR"`. Coudl you explan what is a difference.

dirgim: The SHARED_DIR is used by Prow jobs to share files between different testing steps since there isn't a different mechanism to relay information between steps.
This is obviously not required when running the steps in Ansible so I had decided to point the directories to the same path by default when I updated the script to the version that introduced usage of the SHARED_DIR.
I'm not entirely satisfied with this solution as it looks weird in the ansible output, so will address it in an upcoming PR.
This should not hold up this PR as it's unrelated to it, though.

lslebodn: Thank you for explanation

lslebodn:
>the fix for CI issues is in #143

@p3ck please rebase PR

p3ck:
@dirgim Have I done anything wrong to explain the error for CVP/cvp-redhat-operator-bundle-image-validation-test?

PR #144, opened 2021-04-30
[CVP-1923] Don't fail when alm-examples are not defined

dirgim:
* Remove check for alm-examples from metadata parsing
* Update the CI unit tests according to the new logic
* Change the file permissions in local test playbook to fix local testing

### DIFF
--- a/.github/workflows/build_and_test_upstream_ci_image.yml
+++ b/.github/workflows/build_and_test_upstream_ci_image.yml
@@ -26,8 +26,9 @@ jobs:
         run: |
           podman --version
           podman build -t upstream_ci_image:latest -f Dockerfiles/ci/Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0
-      - name: "Run example operator-metadata through upstream_ci_image --> negative test"
+
+      - name: "Run example operator-metadata with missing alm-examples through upstream_ci_image --> positive test"
         continue-on-error: false
         run: |
           mkdir test_operator_work_dir output_logs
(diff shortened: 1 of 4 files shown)


.github/workflows/build_and_test_upstream_ci_image.yml
```diff
@@ -30,4 +30,4 @@ jobs:
         continue-on-error: false
         run: |
           mkdir test_operator_work_dir output_logs
-          podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_failed_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
+          podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-wrong-channel:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_failed_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
```
lslebodn, line 33: I would keep even test-case for `example-metadata-without-alm-annotations`.
And we need to ensure it passed certification.

dirgim: I changed the test case back to test `example-metadata-without-alm-annotations` as a positive test.
I decided against adding the new test case as there's no support currently for running different tests on different metadata, and adding it in this PR would be out of scope.

wheelerlaw:
lgtm

wheelerlaw (approved):

sonam1412 (approved):
lgtm

.github/workflows/build_and_test_upstream_ci_image.yml
lslebodn: ty

PR #145, opened 2021-05-03
[CVP-1870] Create a wrapper playbook role for the cvp-common-scorecard Prow step

14rcole:
Add a wrapper for the operator bundle scorecard tests to the upstream test playbooks.

This role will be run by DCI after the deployment test is complete


14rcole:
@p3ck PTAL

roles/operator_bundle_scorecard_tests/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+- name: "Run operator scorecard test"
+  block:
+  - name: "Set facts for the Operator groups"
+    set_fact:
+      OO_BUNDLE: "{{ OO_BUNDLE }}"
+
+  # incase of failure the following task fails
+  # we will be ignoring errors to display the results
+  - name: "Generate namespace name if not provided"
+    shell: "{{ lookup('template', 'optera') }}"
```
dirgim, line 11: I'd advise to use the `file` type for lookup and moving the `optional-operators-cvp-common-scorecard-commands.sh` script from  `templates` folder to `files`, since the `template` type could cause issues if there are any Ansible-readable strings in the script. 

Also I'm guessing that `optera` is a typo and should be the script's filename.

optional-operators-subscribe.yml
```diff
@@ -8,3 +8,7 @@
     - name: "Run optional operators subscribe step"
       include_role:
         name: optional-operators-subscribe
+
+    - name: "Run the operator scorecard tests step"
```
dirgim, line 12: We should probably either rename this playbook to make it clear that it's meant to be run by external tests in DCI, or make running this role optional based on an input parameter. The name `optional-operators-subscripte.yml` for this playbook looks a bit confusing otherwise.

dirgim (changes requested):

optional-operators-subscribe.yml
14rcole: Would it be better if I made a separate playbook?  I combined them because the scorecard test has to be run after the subscribe test, but I don't see anything wrong with two separate playbooks

p3ck: That would be fine.  I was planning that it would be a different playbook like I show here:

https://github.com/redhat-cip/ansible-role-dci-cvp/blob/master/tasks/run_cvp_operator.yml

roles/operator_bundle_scorecard_tests/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: "Run operator scorecard test"
+  block:
+  - name: "Set facts for the Operator groups"
+    set_fact:
+      OO_BUNDLE: "{{ OO_BUNDLE }}"
```
p3ck, line 6: Isn't it already set?

roles/operator_bundle_scorecard_tests/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+
+  # incase of success/failure the output is being displayed the
+  - name: "Debug"
+    debug:
+      msg: "{{ shell_script_output.stdout_lines }}"
+
+  always:
+  - name: "Store the result in a JSON file"
+    copy:
+      dest: "optional_operator_subscribe.json"
```
p3ck, line 27: I think we need a full path here.

p3ck (changes requested):

roles/operator_bundle_scorecard_tests/tasks/main.yml
14rcole: Should it go in {{ ARTIFACT_DIR }}?

roles/operator_bundle_scorecard_tests/tasks/main.yml
14rcole: `OO_BUNDLE` is set in Ansible, but not as an environment variable

roles/operator_bundle_scorecard_tests/tasks/main.yml
p3ck: Yes

roles/operator_bundle_scorecard_tests/tasks/main.yml
p3ck: This won't set it as an environment variable though.

14rcole: You're right, I thought this comment was on another line.  Updated

roles/operator_bundle_scorecard_tests/tasks/main.yml
14rcole: Fixed

dirgim:
@14rcole the ansible part of the PR looks good to me, but I'm not seeing the `optional-operators-cvp-common-scorecard-commands.sh` script that should be added to the `files/` dir in the new role.

14rcole:
Fixed

roles/operator_bundle_scorecard_tests/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+
+  # incase of success/failure the output is being displayed the
+  - name: "Debug"
+    debug:
+      msg: "{{ shell_script_output.stdout_lines }}"
+
+  always:
+  - name: "Store the result in a JSON file"
+    copy:
+      dest: "{{ ARTIFACT_DIR }}/operator_bundle_scorecard_tests.json"
```
dirgim, line 23: Missed this before - the shell script output will not be in `json` format - it will be regular text output.

operator-bundle-scorecard-tests.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: "Run optional operators subscribe"
```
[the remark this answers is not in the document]
14rcole: Fixed

dirgim (approved):
lgtm

PR #146, opened 2021-05-04
[CVP-1899]-setting exit status code & error msg

jsztuka:
According to the description, in case of 
Exit code between **50-99** we are letting the user know that the issue occured and where he can find mode explicit description of it.
In case of any other non-zero exit code we save the error message in file as well.
All error messages are saved to files with **.errormessage** extension. Those files are located where the entrypoint script is located.

### DIFF
--- a/.github/workflows/build_and_test_midstream_image.yml
+++ b/.github/workflows/build_and_test_midstream_image.yml
@@ -31,45 +31,9 @@ jobs:
           podman --version
           cd Dockerfiles/midstream/
           podman build -t midstream_image:latest -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0
-      - name: "Run example operator-metadata through midstream ci image  --> positive test"
+
+      - name: "Run unit tests"
         continue-on-error: false
         run: |
-          mkdir ./output_logs
-          podman run -it -v $PWD:/project/operator-test-playbooks -v ./output_logs:/project/output midstream_image:latest
-
-      - name: "Run example operator-metadata through midstream ci image  --> negative parsing test"
-        continue-on-error: true
-        run: |
-          rm -rf ./output_logs
-          mkdir ./output_logs
-          podman run -it -v $PWD:/project/operator-test-playbooks -v ./output_logs:/project/output -e IMAGE_TO_TEST=quay.io/cvpops/test-operator:missing-alm-examples-v1 midstream_image:latest
-
-      - name: "Succeed if the negative test fails"
-        continue-on-error: false
-        if: ${{ failure() }}
-        run: |
-          exit 0
-
-      - name: "Fail if the previous job skips"
-        if: ${{ cancelled() }}
-        run: |
-          exit 0
-
-      - name: "Run example operator-metadata through midstream ci image  --> negative operator-sdk validation test"
-        continue-on-error: true
-        run: |
-          rm -rf ./output_logs
-          mkdir ./output_logs
-          podman run -it -v $PWD:/project/operator-test-playbooks -v ./output_logs:/project/output -e IMAGE_TO_TEST=quay.io/cvpops/test-operator:invalid-dependencies-v1 midstream_image:latest
-
-      - name: "Succeed if the negative test fails"
-        continue-on-error: false
-        if: ${{ failure() }}
-        run: |
-          exit 0
-
-      - name: "Fail if the previous job skips"
-        if: ${{ cancelled() }}
-        run: |
-          exit 0
+          podman run -it -v $PWD:/project/operator-test-playbooks midstream_image:latest python3 /unit_tests.py
 
--- a/Dockerfiles/midstream/README.md
+++ b/Dockerfiles/midstream/README.md
@@ -1,6 +1,14 @@
 # operator-bundle-validate-container
 CVP midstream operator bundle validate container
 
+## Prerequisites
+
+In order to run unit_tests.py on your local without any issues, you need to make sure 
+you run the command with specific options. 
+Here is an example of running the unit_tests.py locally
+`sudo podman run -it -v $PWD:/project/operator-test-playbooks:z <midstream_image> /unit_tests.py`
+First thing you reckon is that you are running this command directly from upstream repo `operator-test-playbooks` and you have build your own midstream_image locally from Dockerfiles/midstream/Dockerfile.
+
 # Usage
 
 ## Building image on local machine 
--- a/Dockerfiles/midstream/run_tests.py
+++ b/Dockerfiles/midstream/run_tests.py
@@ -6,7 +6,8 @@ import unittest
 import subprocess
 from os import path
 
-class RunMidstreamCVPTests(unittest.TestCase):
+
+class RunMidstreamCVPTests():
 
     def setUp(self):
         self.operator_dir = os.getenv('OPERATOR_DIR',
@@ -17,8 +18,7 @@ class RunMidstreamCVPTests(unittest.TestCase):
                                   "/project/output/")
         self.playbooks_dir = os.getenv('PLAYBOOKS_DIR',
                                        "/project/operator-test-playbooks/")
-        self.image_to_test = os.getenv('IMAGE_TO_TEST',
-                                       'quay.io/cvpops/test-operator:v1.0-16')
+        self.image_to_test = os.getenv('IMAGE_TO_TEST')
         self.umoci_bin_path = os.getenv('UMOCI_BIN_PATH',
                                         '/usr/local/bin/umoci')
 
@@ -31,7 +31,7 @@ class RunMidstreamCVPTests(unittest.TestCase):
                                                     operator_work_dir=self.operator_work_dir,
                                                     bundle_image=self.image_to_test,
                                                     umoci_bin_path=self.umoci_bin_path)
-        result = subprocess.run(exec_cmd, shell=True)
+        result = subprocess.run(exec_cmd, shell=True, capture_output=True)
         return result
 
     def run_validate_operator_bundle(self):
@@ -40,26 +40,58 @@ class RunMidstreamCVPTests(unittest.TestCase):
                     -e 'operator_dir={operator_dir}' \
                     -e 'operator_work_dir={operator_work_dir}' \
                     -e 'work_dir={work_dir}'".format(operator_dir=self.operator_dir,
-                                                    operator_work_dir=self.operator_work_dir,
-                                                    work_dir=self.work_dir)
-        result = subprocess.run(exec_cmd, shell=True)
+                                                     operator_work_dir=self.operator_work_dir,
+                                                     work_dir=self.work_dir)
+        result = subprocess.run(exec_cmd, shell=True, capture_output=True)
         return result
 
     def test_for_extract_and_validate_bundle_image(self):
+
+        self.setUp()
+        global exit_code
+        exit_code = 0
+        # check if IMAGE_TO_TEST is defined, return exit_code 102 in case it's not
+        if (self.image_to_test is None):
+            print("Environment variable IMAGE_TO_TEST not set! Stopping the tests.")
+            with open(".errormessage", 'w') as error_msg:
+                print("Result code: 102 Error message: Environment variable IMAGE_TO_TEST not set!", file=error_msg)
+            exit_code = 102
+            return exit_code
+        if (self.operator_dir != '/project/operator_dir/' or
+           self.operator_work_dir != '/project/test_operator_work_dir/' or
+           self.playbooks_dir != '/project/operator-test-playbooks/' or
+           self.umoci_bin_path != '/usr/local/bin/umoci' or
+           self.work_dir != '/project/output/'):
+            print("Environment variable misconfigured!")
+            with open(".errormessage", 'w') as error_msg:
+                print("Result code: 103 Error message: Environment variable was not set correctly!", file=error_msg)
+            exit_code = 103
+            return exit_code
         result = self.run_extract_operator_bundle()
-        self.assertEqual(0,
-                         result.returncode,
-                         "extract-operator-bundle.yml failed")
+        if (result.returncode != 0):
+            print("Ansible playbook extract-operator-bundle.yml failed with result code: %s , see the file .errormessage for more info." % result.returncode)
+            with open(".errormessage", "w") as error_msg:
+                print("Result code: " + str(result.returncode), "Error message: " + str(result.stderr), file=error_msg)
+            exit_code = 50
+            return exit_code
         result = self.run_validate_operator_bundle()
-        self.assertEqual(0,
-                         result.returncode,
-                         "validate-operator-bundle.yml failed")
-        self.assertTrue(path.exists("/project/output/validation-rc.txt"))
-        self.assertTrue(path.exists("/project/output/validation-output.txt"))
-        self.assertTrue(path.exists("/project/output/validation-version.txt"))
+        assert (path.exists("/project/output/validation-rc.txt"))
+        assert (path.exists("/project/output/validation-output.txt"))
+        assert (path.exists("/project/output/validation-version.txt"))
         with open('/project/output/validation-rc.txt', 'r') as reader:
             rc = reader.read()
-            self.assertEqual(0, int(rc), "validation of bundle image failed")
+            if (int(rc) != 0):
+                print("Image bundle validation failed with result code: %s , see /project/output/validation-output.txt file for more info." % int(rc))
+                exit_code = 70
+                return exit_code
+        return 0
 
 if __name__ == '__main__':
-    unittest.main()
+    resultcodes = [0, 50, 70, 102, 103]
+    runTest = RunMidstreamCVPTests()
+    with open("tests_result.log", 'w') as f:
+        if (runTest.test_for_extract_and_validate_bundle_image() not in resultcodes):
+            exit_code = 1
+            print("Error occured during unit tests, please see .errormessage file for more info.")
+            os.rename(r'tests_result.log', r'.errormessage')
+    sys.exit(exit_code)
--- a/Dockerfiles/midstream/unit_tests.py
+++ b/Dockerfiles/midstream/unit_tests.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+import unittest
+import os
+import subprocess
+
+
+class Testing(unittest.TestCase):
+    # Running the container with correctly set environment variables, all tasks from run_tests.py should pass
+    def test_default_positive(self):
+        exec_cmd = """
+        export IMAGE_TO_TEST=quay.io/cvpops/test-operator:v1.0-16
+        /run_tests.py
+        """
+        self.assertFalse(os.path.exists(".errormessage"), "File .errormessage should not be present!")
+        result = subprocess.run(exec_cmd, shell=True)
+        self.assertEqual(0, result.returncode)
+
+    # Set env variable to custom made image that makes playbook extract-operator-bundle.yml fail
+    # Due to parsing error
+    def test_negative_parsing(self):
+        exec_cmd = """
+        export IMAGE_TO_TEST=quay.io/cvpops/test-operator:missing-alm-examples-v1
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open(".errormessage") as error_file:
+            self.assertIn("Result code:", error_file.read(), "Result code not found in %s" % error_file)
+        self.assertEqual(50, result.returncode)
+
+    # Set env variable to custom made image that makes validation fails with following error:
+    # ERRO[0003] error validating format in /tmp/bundle-688328851: Bundle validation errors: couldn't
+    # parse dependency of type olm.crd
+    def test_negative_image_bundle_validation(self):
+        exec_cmd = """
+        export IMAGE_TO_TEST=quay.io/cvpops/test-operator:invalid-dependencies-v1
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open("/project/output/validation-output.txt") as error_file:
+            self.assertIn("Bundle validation errors: couldn't parse dependency of type olm.crd", error_file.read(), "Result code not found in %s" % error_file)
+        self.assertEqual(70, result.returncode)
+
+    # Don't set env variable IMAGE_TO_TEST case
+    def test_negative_image_to_test_not_set(self):
+        exec_cmd = """
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open(".errormessage") as error_file:
+            self.assertIn("Result code: 102 Error message: Environment variable IMAGE_TO_TEST not set!", error_file.read(), "Result code not found in %s" % error_file)
+        self.assertEqual(102, result.returncode)
+
+    # Set incorrect env variable UMOCI_BIN_PATH for ansible playbook extract-operator-bundle.yml
+    def test_negative_extract_operator_bundle(self):
+        exec_cmd = """
+        export IMAGE_TO_TEST=quay.io/cvpops/test-operator:v1.0-16
+        export UMOCI_BIN_PATH=/root
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open(".errormessage") as error_file:
+            self.assertIn("Result code: 103 Error message: Environment variable was not set correctly!", error_file.read(), "Result code not found in %s" % error_file)
+        self.assertEqual(103, result.returncode)
+
+if __name__ == '__main__':
+    unittest.main()
--- a/Dockerfiles/midstream/Dockerfile
+++ b/Dockerfiles/midstream/Dockerfile
@@ -14,4 +14,5 @@ RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm6
     dnf install --setopt=install_weak_deps=False -y git-core ansible skopeo && \
     dnf clean all
 ADD ./run_tests.py /run_tests.py
+ADD ./unit_tests.py /unit_tests.py
 CMD ["/run_tests.py"]


Dockerfiles/midstream/run_tests.py
```diff
@@ -42,18 +42,26 @@ def run_validate_operator_bundle(self):
...
-        result = subprocess.run(exec_cmd, shell=True)
+        result = subprocess.run(exec_cmd, shell=True, capture_output=True)
         return result
 
     def test_for_extract_and_validate_bundle_image(self):
         result = self.run_extract_operator_bundle()
-        self.assertEqual(0,
-                         result.returncode,
-                         "extract-operator-bundle.yml failed")
+        if(50 <= result.returncode <= 99):
```
lslebodn, line 50: ansible playbooks does not return such error codes.

This scrpt `run_test.py` has wto retun such error codes.

lslebodn (approved):
There are missing tests for return codes returned from the script `run_test.py`

lslebodn:
Note about testing return codes which need to be implemented in this PR.
ATM we use github actions as a "testing" framework. It is wrong.

We need to be able to test changes also locally.
The current implementation of the file `Dockerfiles/midstream/run_tests.py` uses  python module `unittest`.
It might work in some cases but it is far away from ideal way of writing wrapper script. (you cannot control return codes for assertions.

But we can use the python module `unittest` for writing tests for the script run_tests.py and it can be packaged in the same container images so we can run the same tests in github action and also locally

jsztuka:
Modified the script so the run_tests.py returns exit code according to the issue that occured. Now we should write unit tests for the run_tests.py result codes.
Do we want it as a part of this PR or we want to create a new one?

Dockerfiles/midstream/run_tests.py
```diff
@@ -42,18 +42,26 @@ def run_validate_operator_bundle(self):
                     -e 'work_dir={work_dir}'".format(operator_dir=self.operator_dir,
                                                     operator_work_dir=self.operator_work_dir,
                                                     work_dir=self.work_dir)
-        result = subprocess.run(exec_cmd, shell=True)
+        result = subprocess.run(exec_cmd, shell=True, capture_output=True)
         return result
 
     def test_for_extract_and_validate_bundle_image(self):
+        global exitCode
```
samvarankashyap, line 49: I would discourage usage of CamelCase inside Python. 
it should be snake casing. 
@jsztuka

jsztuka: Ok, will change that.

jsztuka: changed

.github/workflows/build_and_run_unit_tests.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      # Runs a set of commands using the runners shell
+      - name: "Build the midstream image inside Dockerfiles"
+        run: |
+          podman --version
+          cd Dockerfiles/midstream/
+          podman build -t midstream_image:latest -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0
```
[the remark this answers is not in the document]
samvarankashyap: +1

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
+
+import unittest
+import os
+import subprocess
+
+
+class Testing(unittest.TestCase):
```
samvarankashyap, line 8: Use pep8 styling for this file. @jsztuka

samvarankashyap (commented):
Looks fine for me. 
except for the additional file mentioned by lukas

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
```
lslebodn, line 1: I would include file into container image.

That way you needn;t build anything because you have all necessary dependencies already in midstream image including `run_tests.py` which you need to test.

And for running tests you will just use podman -ti --rm run  $name_of_midstream_image python3 unit_tests.py 

It will simplify the unit test itself as well. You will be able to get rid of all podman login.

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,43 @@
...
+        subprocess.run(exec_cmd, shell=True)
+
+    def test_postive(self):
+        exec_cmd = """
+        podman run -it -v ../../:/project/operator-test-playbooks -v ../.././output_logs:/project/output midstream:latest
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        self.assertEqual(0,result.returncode)
+
+    def test_negative_55(self):
```
lslebodn, line 21: Do not use numbers in the name, but rather use name of test-case which you want to cover.
e.g. extract operator bundle failed, operator sdk failed ...

Dockerfiles/midstream/run_tests.py
```diff
@@ -42,24 +42,43 @@ def run_validate_operator_bundle(self):
...
+                exit_code = 70
+                return exit_code
+            #self.assertEqual(0, int(rc), "validation of bundle image failed")
 
 if __name__ == '__main__':
-    unittest.main()
+    tests_result_file = open('tests_result.log','w')
+    unittest.main(testRunner=unittest.TextTestRunner(stream=tests_result_file), exit=False)
+    with open("tests_result.log") as f:
+        if 'F' in f.read():
```
lslebodn, line 80: Please do not rely on the unittest module here.

We just need simple python script which will execute two ansible playbooks and set expected return codes.
And do something with log files.

Using the unitest module just make things more complicated.

jsztuka: This section handles the case in which the unit tests fails on any other occasion than the previous ones, and the **exit=False** needs to be there if we want to send custom return codes.

jsztuka: In case we change it to basic unittest.main() method, we would not be ablo to do two things:
send custom return codes, handle errors besides those between 50-99.

lslebodn: One more time.
There is not any benefit of using the unitest module for wrapper script. The official documentation says:
```
The unittest unit testing framework was originally inspired by JUnit and has a similar flavor as major unit testing frameworks in other languages.
```
https://docs.python.org/3/library/unittest.html

We do not do any unit testing in the file `run_tests.py` It is just a wrapper script for ansible playbooks because we cannot return different return code from ansible paybook.

@samvarankashyap  did very wrong decision in the initial implementation. But he was sick and we wanted to merge patches to unblock other team.

The usage of the unittest module just makes everything more complicated.
* You have to use global variables. 
* "parse" output from test runner

It will be much simpler without the usage of unittest module

lslebodn: @jsztuka PLease focus mainly on requested changes to `unit_tests.py` 

We should firstly focus on tests and than on refactoring ( and it can be done even later together with CVP-1928: Midstream-CVP container yields tests results in a JUnit XML format

Dockerfiles/midstream/run_tests.py
```diff
@@ -42,24 +42,43 @@ def run_validate_operator_bundle(self):
...
         self.assertTrue(path.exists("/project/output/validation-output.txt"))
         self.assertTrue(path.exists("/project/output/validation-version.txt"))
         with open('/project/output/validation-rc.txt', 'r') as reader:
             rc = reader.read()
-            self.assertEqual(0, int(rc), "validation of bundle image failed")
+            if(int(rc) != 0):
+                print("Image bundle validation failed with result code: %s , see /project/output/validation-output.txt file for more info." % int(rc))
+                exit_code = 70
+                return exit_code
+            #self.assertEqual(0, int(rc), "validation of bundle image failed")
```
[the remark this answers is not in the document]
jsztuka: Removed.

Dockerfiles/midstream/unit_tests.py
jsztuka: Tests renamed.

Dockerfiles/midstream/unit_tests.py
jsztuka: Unit tests are now part of container image.

.github/workflows/build_and_run_unit_tests.yml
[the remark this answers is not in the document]
jsztuka: changed

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_postive(self):
+        exec_cmd = """
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        self.assertEqual(0, result.returncode)
+
+    def test_negative_extract_operator_bundle(self):
+        exec_cmd = """
+        export UMOCI_BIN_PATH=/root
```
lslebodn, line 18: Please add commemnt to the function what do you try to test here.
It is not clear to me why we should fail with return code 55 for invalid umoci path

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,41 @@
...
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        self.assertEqual(55, result.returncode)
+
+    def test_negative_validate_operator_bundle(self):
+        exec_cmd = """
+        export WORK_DIR=/kandel
```
lslebodn, line 26: Please document why we should fail with return code 60 here.
Is it because of wrong work directory ?

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,41 @@
...
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        self.assertEqual(60, result.returncode)
+
+    def test_negative_image_bundle_validation(self):
+        exec_cmd = """
+        export IMAGE_TO_TEST=quay.io/cvpops/test-operator:invalid-dependencies-v1
+        /run_tests.py
```
lslebodn, line 35: please elaborate what is wrong with expected IMAGE_TO_TEST.

lslebodn:
@jsztuka we shoudl also check expected strings in the generated file `.errormessage`

.github/workflows/build_and_test_midstream_image.yml
```diff
@@ -31,9 +31,17 @@ jobs:
...
+        run: |
+          mkdir ./output_logs
+          podman run -it -v $PWD:/project/operator-test-playbooks -v ./output_logs:/project/output midstream_image:latest python3 /unit_tests.py
+
       - name: "Run example operator-metadata through midstream ci image  --> positive test"
         continue-on-error: false
         run: |
+          rm -rf ./output_logs
           mkdir ./output_logs
           podman run -it -v $PWD:/project/operator-test-playbooks -v ./output_logs:/project/output midstream_image:latest
```
lslebodn, line 46: We do not need anymore following stages.
* "Run example operator-metadata through midstream ci image  --> negative parsing test"
* "Run example operator-metadata through midstream ci image  --> negative operator-sdk validation test"
Theys should be already covered by your file and they do not check return code anyway

Dockerfiles/midstream/unit_tests.py
jsztuka: Yes exactly, I will create comments per unit test so it is clear.

jsztuka:
> @jsztuka we shoudl also check expected strings in the generated file `.errormessage`

Added file content check for those two files we generate text for. Is that sufficient enought?

.github/workflows/build_and_test_midstream_image.yml
jsztuka: Removed.

jsztuka:
This is the issue that is causing metadata validation to fail:
https://pulp-docker.corp.redhat.com/pulp/api/v2/repositories/redhat-e2e-container-e2e-container-test-product-2/?distributors=True

jsztuka:
Failure is not connected to the changes in the PR.

jsztuka:
I don't see failed shared-lib-tests in the CVP-CI, retriggering test.

.github/workflows/build_and_test_midstream_image.yml
lslebodn: Could you explain why why do we skill the stage "Run example operator-metadata through midstream ci image  --> positive test"

jsztuka: Removed redundant test that is already present in unit tests.

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,54 @@
...
+        exec_cmd = """
+        export UMOCI_BIN_PATH=/root
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open("extract-operator-bundle.errormessage") as error_file:
+            if "Result code:" not in error_file.read():
+                print("Result code not found in %s" % error_file)
+                assert False
+        self.assertEqual(55, result.returncode)
```
lslebodn, line 27: Based on the description this should not be covered by return codes 50-99.
But with "Other Codes" because wrong env variable means either broken default in container image
or wrong overwrite when running the container.

In another words, such test case is exactly "a system or an infrastructure issue"  which should be covered by other return codes based on requirements in the story.

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,54 @@
...
+        exec_cmd = """
+        export WORK_DIR=/kandel
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open("validate-operator-bundle.errormessage") as error_file:
+            if "Result code:" not in error_file.read():
+                print("Result code not found in %s" % error_file)
+                assert False
+        self.assertEqual(60, result.returncode)
```
lslebodn, line 40: Based on the description this should not be covered by return codes 50-99.
But with "Other Codes" because wrong env variable for workdirectory means either broken default in container image
or wrong overwrite when running the container.

In another words, such test case is exactly "a system or an infrastructure issue"  which should be covered by other return codes based on requirements in the story.

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,54 @@
...
+        with open("extract-operator-bundle.errormessage") as error_file:
+            if "Result code:" not in error_file.read():
+                print("Result code not found in %s" % error_file)
+                assert False
+        self.assertEqual(55, result.returncode)
+        
+    # Set incorrect env variable WORK_DIR for ansible playbook validate-operator-bundle.yml - playbook fails to validate
+    def test_negative_validate_operator_bundle(self):
+        exec_cmd = """
+        export WORK_DIR=/kandel
```
lslebodn, line 32: `kandel` -> `non-existing-directory` :-)

lslebodn (changes requested):

Dockerfiles/midstream/unit_tests.py
jsztuka: Correct, will rename it to be more understandable.

lslebodn:
> Added file content check for those two files we generate text for. Is that sufficient enough?

We should test as much as possible. Because everything what we test will be guaranteed to be stable for midstream team

Dockerfiles/midstream/unit_tests.py
jsztuka: Changed to 102.

Dockerfiles/midstream/unit_tests.py
jsztuka: Changed to 103

jsztuka:
> > Added file content check for those two files we generate text for. Is that sufficient enough?
> 
> We should test as much as possible. Because everything what we test will be guaranteed to be stable for midstream team

Added one more test for file content.

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+import unittest
+import os
+import subprocess
+
+
+class Testing(unittest.TestCase):
+    # Running the container with correctly set environment variables, all tasks from run_tests.py should pass
+    def test_postive(self):
+        exec_cmd = """
+        /run_tests.py
```
lslebodn, line 11: I know @samvarankashyap introduced such behaviour but IMHO it is invalid case when env IMAGE_TO_TEST
is not specified. Using default image for testing is wrong.
People could do a typo in env variable IMAGE_TO_TEST and we would test default image which will pass all checks. and they can assume their image is good.

We have to fail with error for missing env variable IMAGE_TO_TEST

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        self.assertEqual(0, result.returncode)
```
lslebodn, line 14: Could you explain why we do not check generated file? `.errormessage` ?

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_negative_validate_operator_bundle(self):
+        exec_cmd = """
+        export WORK_DIR=/non_existing_dir
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open("validate-operator-bundle.errormessage") as error_file:
+            if "Result code:" not in error_file.read():
+                print("Result code not found in %s" % error_file)
+                assert False
```
lslebodn, line 39: Could you explain why you do not use self.assert.* function ?

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+        
+    # Set incorrect env variable WORK_DIR for ansible playbook validate-operator-bundle.yml - playbook fails to validate
+    def test_negative_validate_operator_bundle(self):
+        exec_cmd = """
+        export WORK_DIR=/non_existing_dir
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open("validate-operator-bundle.errormessage") as error_file:
+            if "Result code:" not in error_file.read():
```
lslebodn, line 37: Can we change this if to assert ?

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # Set incorrect env variable UMOCI_BIN_PATH for ansible playbook extract-operator-bundle.yml - playbook fails to extract
+    def test_negative_extract_operator_bundle(self):
+        exec_cmd = """
+        export UMOCI_BIN_PATH=/root
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open("extract-operator-bundle.errormessage") as error_file:
+            if "Result code:" not in error_file.read():
```
lslebodn, line 24: Can we change if to assert ? (self.assert.*)

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+    # ERRO[0003] error validating format in /tmp/bundle-688328851: Bundle validation errors: couldn't
+    # parse dependency of type olm.crd
+    def test_negative_image_bundle_validation(self):
+        exec_cmd = """
+        export IMAGE_TO_TEST=quay.io/cvpops/test-operator:invalid-dependencies-v1
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open("/project/output/validation-output.txt") as error_file:
+            if "Bundle validation errors: couldn't parse dependency of type olm.crd" not in error_file.read():
```
lslebodn, line 52: Can we change if to assert ? (self.assert.*)

Dockerfiles/midstream/run_tests.py
```diff
@@ -42,24 +42,45 @@ def run_validate_operator_bundle(self):
...
+        exit_code = 0
         result = self.run_extract_operator_bundle()
-        self.assertEqual(0,
-                         result.returncode,
-                         "extract-operator-bundle.yml failed")
+        if(result.returncode != 0):
+            print("Ansible playbook extract-operator-bundle.yml failed with result code: %s , see the file extract-operator-bundle.errormessage for more info." % result.returncode)
+            with open("extract-operator-bundle.errormessage", "w") as errorMsg:
+                print("Result code: " + str(result.returncode), "Error message: " + str(result.stderr), file=errorMsg)
+            exit_code = 102
```
lslebodn, line 58: IIUC we will always return exit code 102 for any failure detected by the playbook extract-operator-bundle.yml.
But requirements says we should return 50-99.

We need to return exit code 102 just for invalid "input" env variable

Dockerfiles/midstream/run_tests.py
```diff
@@ -42,24 +42,45 @@ def run_validate_operator_bundle(self):
...
+        
+        self.setUp()
+        global exit_code
+        exit_code = 0
         result = self.run_extract_operator_bundle()
-        self.assertEqual(0,
-                         result.returncode,
-                         "extract-operator-bundle.yml failed")
+        if(result.returncode != 0):
+            print("Ansible playbook extract-operator-bundle.yml failed with result code: %s , see the file extract-operator-bundle.errormessage for more info." % result.returncode)
```
lslebodn, line 55: The requirements in the story says:
```
.errormessage 
Should be place in $PWD (where the entrypoint script is located).
```
The requirements dows not say anything about `extract-operator-bundle.errormessage`

Dockerfiles/midstream/run_tests.py
```diff
@@ -42,24 +42,45 @@ def run_validate_operator_bundle(self):
...
+            return exit_code
         result = self.run_validate_operator_bundle()
-        self.assertEqual(0,
-                         result.returncode,
-                         "validate-operator-bundle.yml failed")
-        self.assertTrue(path.exists("/project/output/validation-rc.txt"))
-        self.assertTrue(path.exists("/project/output/validation-output.txt"))
-        self.assertTrue(path.exists("/project/output/validation-version.txt"))
+        if(result.returncode != 0):
+            print("Ansible playbook validate-operator-bundle.yml failed with result code: %s , see the file validate-operator-bundle.errormessage for more info." % result.returncode)
```
lslebodn, line 62: The requirements in the story says:
```
.errormessage 
Should be place in $PWD (where the entrypoint script is located).
```
The requirements dows not say anything about `validate-operator-bundle.errormessage`

Dockerfiles/midstream/run_tests.py
```diff
@@ -42,24 +42,45 @@ def run_validate_operator_bundle(self):
...
 
 if __name__ == '__main__':
-    unittest.main()
+    resultcodes = [0,102,103,70]
+    runTest = RunMidstreamCVPTests()
+    with open("tests_result.log",'w') as f:
+        if (runTest.test_for_extract_and_validate_bundle_image() not in resultcodes):
+            exit_code = 1
+            print("Error occured during unit tests, please see tests_result.errormessage file for more info.")
+            os.rename(r'tests_result.log',r'tests_result.errormessage')
```
lslebodn, line 85: The requirements in the story says:
```
.errormessage 
Should be place in $PWD (where the entrypoint script is located).
```
The requirements does not say anything about tests_result.errormessage

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,59 @@
...
+                print("Result code not found in %s" % error_file)
+                assert False
+        
+        self.assertEqual(70, result.returncode)
+
+if __name__ == '__main__':
+    unittest.main()
```
lslebodn, line 59: I miss the test-case for `quay.io/cvpops/test-operator:missing-alm-examples-v1`
It was in the `.github/workflows/build_and_test_midstream_image.yml`

Dockerfiles/midstream/unit_tests.py
lslebodn: I would recommend to explicitly pass good image here.
And write new test-case to fail with error (Failure due to what is probably  a system or an infrastructure issue)
for missing env variable `IMAGE_TO_TEST`

Dockerfiles/midstream/unit_tests.py
jsztuka: Will use self.assertTrue(), I would say it is the best fit here.

Dockerfiles/midstream/unit_tests.py
jsztuka: Will use self.assertTrue(), I would say it is the best fit here.

Dockerfiles/midstream/unit_tests.py
jsztuka: Will use self.assertTrue(), I would say it is the best fit here.

Dockerfiles/midstream/unit_tests.py
jsztuka: Will use self.assertTrue(), I would say it is the best fit here.

Dockerfiles/midstream/unit_tests.py
lslebodn: There is also `assertIn`
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn

Dockerfiles/midstream/unit_tests.py
lslebodn: There is also assertIn
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn

Dockerfiles/midstream/unit_tests.py
lslebodn: There is also assertIn
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIn

Dockerfiles/midstream/run_tests.py
jsztuka: Changed to .errormessage

Dockerfiles/midstream/run_tests.py
jsztuka: Changed to .errormessage

Dockerfiles/midstream/run_tests.py
jsztuka: Changed to .errormessage

Dockerfiles/midstream/unit_tests.py
jsztuka: I removed it due to comment:
We do not need anymore following stages.

"Run example operator-metadata through midstream ci image --> negative parsing test"
"Run example operator-metadata through midstream ci image --> negative operator-sdk validation test"
Theys should be already covered by your file and they do not check return code anyway

Didn't realized that I needed to implement it in Unit tests, sorry my bad.

Dockerfiles/midstream/unit_tests.py
jsztuka: Check added.

Dockerfiles/midstream/unit_tests.py
jsztuka: Refactored, added.

Dockerfiles/midstream/run_tests.py
lslebodn: @jsztuka I still cannot see the change.

Dockerfiles/midstream/run_tests.py
lslebodn: @jsztuka I still cannot see the change.

Dockerfiles/midstream/run_tests.py
lslebodn: @jsztuka I still cannot see the change.

Dockerfiles/midstream/unit_tests.py
lslebodn: @jsztuka I still cannot see the change.

Dockerfiles/midstream/unit_tests.py
lslebodn: @jsztuka I still cannot see the change.

Dockerfiles/midstream/run_tests.py
jsztuka: Currently I don't know of a different way to simulate test : test_negative_validate_operator_bundle()
Other then setting wrong work dir.

jsztuka: We need to work on image that pass extraction but fails on validation, we currently don't correspond with such image.

Dockerfiles/midstream/unit_tests.py
jsztuka: Desn't improve readability or anything in code in my opinion, but implemented.

Dockerfiles/midstream/run_tests.py
dirgim: @jsztuka let me know if you need help creating an image that satisfies these requirements. The quay.io/cvpops/test-operator:invalid-dependencies-v1 image passes extraction and fails on the operator-bundle-validate so seems like a good candidate for starters.

jsztuka: Yes I need a help creating the image for that requirements, I should rephrase my sentence. We need container that fails running the playbook: validate-operator-bundle.yml.
The image quay.io/cvpops/test-operator:invalid-dependencies-v1 passes both playbooks.

jsztuka: I currently don't have other option than setting wrong variable name specific only for the playbook. And that does not cover the case that is described in story.

jsztuka: So after shord investigation the only way to fail the ansible playbook on purpose would be just providing incorrect variables before we run the playbook. So we already have it for setting up the non-existend folder for the work_dir. I think this test can be left as it is.

jsztuka: Do we agree on leaving it the way we currently have? Specifying wrong env variable that is used only for the one of the playbooks "WORK_DIR" since there is no other way of simulating such situation.

Dockerfiles/midstream/unit_tests.py
lslebodn: I cannot see change for 103

Dockerfiles/midstream/run_tests.py
```diff
@@ -42,24 +41,61 @@ def run_validate_operator_bundle(self):
...
+        result = subprocess.run(exec_cmd, shell=True, capture_output=True)
         return result
 
     def test_for_extract_and_validate_bundle_image(self):
+
+        self.setUp()
+        global exit_code
+        exit_code = 0
+        #check if IMAGE_TO_TEST is defined, return exit_code 85 in case it's not
+        if(self.image_to_test == None):
```
lslebodn, line 53: Python expect space after if  in pep8.

Please also fix most obvious pep8 violation on other places as well.
I would ignore line too long for now.

```
sh$ git log -p -2 | pycodestyle-3 --diff | grep -v "E501 line too long"
./Dockerfiles/midstream/run_tests.py:9:1: E302 expected 2 blank lines, found 1
./Dockerfiles/midstream/run_tests.py:42:53: E128 continuation line under-indented for visual indent
./Dockerfiles/midstream/run_tests.py:43:53: E128 continuation line under-indented for visual indent
./Dockerfiles/midstream/run_tests.py:52:9: E265 block comment should start with '# '
./Dockerfiles/midstream/run_tests.py:53:31: E711 comparison to None should be 'if cond is None:'
./Dockerfiles/midstream/run_tests.py:59:60: W291 trailing whitespace
./Dockerfiles/midstream/run_tests.py:60:75: W291 trailing whitespace
./Dockerfiles/midstream/run_tests.py:93:1: E305 expected 2 blank lines after class or function definition, found 1
./Dockerfiles/midstream/run_tests.py:94:21: E231 missing whitespace after ','
./Dockerfiles/midstream/run_tests.py:94:24: E231 missing whitespace after ','
./Dockerfiles/midstream/run_tests.py:94:27: E231 missing whitespace after ','
./Dockerfiles/midstream/run_tests.py:94:30: E231 missing whitespace after ','
./Dockerfiles/midstream/run_tests.py:94:34: E231 missing whitespace after ','
./Dockerfiles/midstream/run_tests.py:96:33: E231 missing whitespace after ','
./Dockerfiles/midstream/run_tests.py:100:42: E231 missing whitespace after ','
./Dockerfiles/midstream/unit_tests.py:29:1: W293 blank line contains whitespace
./Dockerfiles/midstream/unit_tests.py:54:1: W293 blank line contains whitespace
./Dockerfiles/midstream/unit_tests.py:66:1: W293 blank line contains whitespace
./Dockerfiles/midstream/unit_tests.py:77:1: E305 expected 2 blank lines after class or function definition, found 1
```

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,78 @@
...
+    
+    # Don't set env variable IMAGE_TO_TEST case
+    def test_negative_image_to_test_not_set(self):
+        exec_cmd = """
+        /run_tests.py
+        """
+        result = subprocess.run(exec_cmd, shell=True)
+        with open(".errormessage") as error_file:
+            self.assertIn("Result code: 102 Error message: Environment variable IMAGE_TO_TEST not set!", error_file.read(), "Result code not found in %s" % error_file)
+        self.assertEqual(102, result.returncode)
```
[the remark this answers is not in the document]
jsztuka: Changed the orded so the return codes are in ascending order.

Dockerfiles/midstream/run_tests.py
jsztuka: Run through pep8, changed code accordingly.

Dockerfiles/midstream/unit_tests.py
jsztuka: So I presume changing this one to 100+ or "other codes" is required.

eisraeli (changes requested):

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,78 @@
+
```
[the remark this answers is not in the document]
jsztuka: Sure thing :)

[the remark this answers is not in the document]
jsztuka: Done.

Dockerfiles/midstream/unit_tests.py
lslebodn: Sorry I overlooked that.
Yes, wrong env variable `WORK_DIR` is not a problem with tested bundle. It is an infrastructure issue.

jsztuka: Added WORK_DIR to decision logic, it is now corresponding to return code 103 if this env var is misconfigured.

Dockerfiles/midstream/README.md
```diff
@@ -1,6 +1,12 @@
 # operator-bundle-validate-container
 CVP midstream operator bundle validate container
 
+## Prerequisites
+
+In order to run unit_tests.py on your local without any issues, you need to make sure your SELinux is set to `Permissive`.
+To chec your current status of SELinux just run command `# getenforce`.
```
dirgim, line 7: Typo: chec -> check

Dockerfiles/midstream/run_tests.py
dirgim: I think we can resolve this conversation based on the code changes that were made.

Dockerfiles/midstream/README.md
jsztuka: Updated, thanks.

dirgim (approved):
LGTM

Dockerfiles/midstream/README.md
```diff
@@ -1,6 +1,12 @@
 # operator-bundle-validate-container
 CVP midstream operator bundle validate container
 
+## Prerequisites
+
+In order to run unit_tests.py on your local without any issues, you need to make sure your SELinux is set to `Permissive`.
+To check your current status of SELinux just run command `# getenforce`.
+To achieve Permissive mode you need to run following command `# setenforce 0`
```
lslebodn, line 8: Please remove such wrong recommendation. There are more and less dangerous ways how to treat SELinux violation.
* Fix SELinuxl labeling on host
* disable SELinux enforcement just for the container

lslebodn (changes requested):

Dockerfiles/midstream/README.md
jsztuka: Changed, specified command to run on the local to successfully run the unit_tests.py

.github/workflows/build_and_test_midstream_image.yml
```diff
@@ -31,45 +31,10 @@ jobs:
...
-        continue-on-error: false
-        if: ${{ failure() }}
-        run: |
-          exit 0
-
-      - name: "Fail if the previous job skips"
-        if: ${{ cancelled() }}
-        run: |
-          exit 0
+          podman run -it -v $PWD:/project/operator-test-playbooks -v ./output_logs:/project/output midstream_image:latest python3 /unit_tests.py
```
lslebodn, line 39: @samvarankashyap  Why do we need to bind-mount playbooks into container image?
It means people in midstream will have to checkout playbooks as well. Which is not optimal solution.
They should have everything necessary in midstream image.

.github/workflows/build_and_test_midstream_image.yml
```diff
@@ -31,45 +31,10 @@ jobs:
...
-      - name: "Fail if the previous job skips"
-        if: ${{ cancelled() }}
         run: |
-          exit 0
-
-      - name: "Run example operator-metadata through midstream ci image  --> negative operator-sdk validation test"
-        continue-on-error: true
-        run: |
-          rm -rf ./output_logs
           mkdir ./output_logs
```
lslebodn, line 38: @jsztuka please remove this line and also  `-v ./output_logs:/project/output` from podman run.
Such part is unused and just caused issues with SELinux due to following AVC:
```
time->Mon May 31 09:47:37 2021
type=AVC msg=audit(1622468857.005:522): avc:  denied  { relabelto } for  pid=23854 comm="python3" name=".ansible_tmppbp1jnxevalidation-version.txt" dev="dm-0" ino=1582971 scontext=system_u:system_r:container_t:s0:c219,c502 tcontext=system_u:object_r:fusefs_t:s0 tclass=file permissive=0
```

Dockerfiles/midstream/unit_tests.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+import unittest
+import os
+import subprocess
+
+
+class Testing(unittest.TestCase):
+    # Running the container with correctly set environment variables, all tasks from run_tests.py should pass
+    def test_default_postive(self):
```
lslebodn, line 9: typo in postive

.github/workflows/build_and_test_midstream_image.yml
jsztuka: Removed both.

.github/workflows/build_and_test_midstream_image.yml
lslebodn: @samvarankashyap  bump

samvarankashyap: @lslebodn Since, we are currently checking out the code from PR. We would like to add the latest changes in PR to be mounted on to the container . The container would be able to access the latest changes in PR while running the tests.

lslebodn: @samvarankashyap  Such use-case might work for pull-request testing.
But it will not work in production. Because we would delegate "checkout" of correct branch to midstream team.
And I do not remember such requirements from midstream team. They just want to use image and choose version of checks based on tag.

samvarankashyap: @lslebodn This is useful for the building the latest and greatest. everytime when a PR is merged. 
However, There needs to be a step which builds on every new tag of operator-test-playbooks. 
Midstream image will have both latest and tagged versions of the images at anypoint. 
Latest being based on every latest commit. 
Does that make sense ? 
@jsztuka Do we have a github action that builds on every tag ?

jsztuka: I'm not sure if on every tag, but on each pull request. I'm not sure if I follow, what do you mean by every tag?

samvarankashyap: @jsztuka  Looks like its already there 
https://github.com/redhat-operator-ecosystem/operator-test-playbooks/blob/master/.github/workflows/build_and_push_image_on_tag.yml#L39 
building new image and pushing it on everytag 
we need to change it to the official image

lslebodn: @samvarankashyap the problem is not with pushing image to quay.

The problem is with the building of the image. The current version of Dockerfile in PR  https://github.com/redhat-operator-ecosystem/operator-test-playbooks/blob/47bc1b952eacdb4c3ce4c9980f956a4f8450c985/Dockerfiles/midstream/Dockerfile
does not contain roles and playbooks inside the image. Which means that midstream people will have to pull image from quay using tag and also checkout playbooks with the same tag.

And Jiri did not change any logic in the Dockerfile. He just add unit test there. Please file a bug for that. We have to fix it ASAP but it should not block this PR

jsztuka: Changed to the official image, can you please confirm this address.

.github/workflows/build_and_push_image_on_tag.yml
```diff
@@ -36,4 +36,4 @@ jobs:
           cd Dockerfiles/midstream/
           podman build -t midstream_image:$latestTag -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=$latestTag
           podman login -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io
-          podman push midstream_image:$latestTag quay.io/samvarankashyap/cvp-midstream-image:$latestTag
```
lslebodn, line 39: @jsztuka Do we have secrets to push there?

jsztuka: Agreed, changed back to original value. New PR should be created regarding this change.

lslebodn (approved):
Lets merge it and track additional changes bug-fixes in different PR/issues

jsztuka:
Failed tests are caused by timeouts, cant reach brew, probably due to ongoing outage:
https://one.redhat.com/legacy/outages/list/details/6070a7981906382bbe2e1b05

jsztuka:
There were some recommendations raised during reviewe meeting with Barak Korren which will be summarized in corresponding Jira story ticket.

PR #149, opened 2021-05-27
[CVP-1980] Paramaeterize scorecard configuration file and authentication

14rcole:
Update scorecard test ansible role and shell script to be compatible with external tests in DCI

### DIFF
--- a/roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
+++ b/roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+# this file is to be kept up-to-date with the one in the openshift/release repo,
+# as documented in the readme for this role
 
 set -o nounset
 set -o errexit
@@ -8,6 +10,8 @@ set -o pipefail
 # only happen in rehearsals. Production jobs should always set the OO_* variable.
 REHEARSAL_BUNDLE="brew.registry.redhat.io/rh-osbs-stage/e2e-e2e-test-operator-bundle-container:8.0-3"
 OO_BUNDLE="${OO_BUNDLE:-$REHEARSAL_BUNDLE}"
+OPENSHIFT_AUTH="${OPENSHIFT_AUTH:-/var/run/brew-pullsecret/.dockerconfigjson}"
+SCORECARD_CONFIG="${SCORECARD_CONFIG:-/tmp/config/scorecard-basic-config.yml}"
 
 # Steps for running the basic operator-sdk scorecard test
 # Expects the standard Prow environment variables to be set and
@@ -15,21 +19,27 @@ OO_BUNDLE="${OO_BUNDLE:-$REHEARSAL_BUNDLE}"
 
 NAMESPACE=$(grep "install_namespace:" "${SHARED_DIR}"/oo_deployment_details.yaml | cut -d ':' -f2 | xargs)
 
-OPERATOR_DIR="${ARTIFACT_DIR}/test-operator-basic"
+pushd "${ARTIFACT_DIR}"
+OPERATOR_DIR="test-operator-basic"
 
 echo "Starting the basic operator-sdk scorecard test for ${OO_BUNDLE}"
 
 echo "Extracting the operator bundle image into the operator directory"
 mkdir -p "${OPERATOR_DIR}"
 pushd "${OPERATOR_DIR}"
-oc image extract "${OO_BUNDLE}" --confirm -a /var/run/brew-pullsecret/.dockerconfigjson
+oc image extract "${OO_BUNDLE}" --confirm -a "${OPENSHIFT_AUTH}"
+# the `oc image extract` command sets the permissions of the files to 0600.  DCI
+# runs the scorecard tests in a container, and thus needs read permission for
+# the container user, which does not own the files.
+chmod -R go+r ./
 popd
 echo "Extracted the following bundle data:"
 tree "${OPERATOR_DIR}"
 
 echo "Running the operator-sdk scorecard test using the basic configuration, json output and storing it in the artifacts directory"
-operator-sdk scorecard --config /tmp/config/scorecard-basic-config.yml \
+operator-sdk scorecard --config "${SCORECARD_CONFIG}" \
                        --namespace "${NAMESPACE}" \
                        --kubeconfig "${KUBECONFIG}" \
                        --output json \
                        "${OPERATOR_DIR}" > "${ARTIFACT_DIR}"/scorecard-output-basic.json || true
+popd
--- a/roles/operator_bundle_scorecard_tests/files/scorecard-basic-config.yml
+++ b/roles/operator_bundle_scorecard_tests/files/scorecard-basic-config.yml
@@ -1,0 +1,23 @@
+# this file is to be kept up-to-date with the one in the operator-scorecard-test-container,
+# as documented in the readme for this role
+kind: Configuration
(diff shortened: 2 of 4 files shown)


dirgim (approved):
LGTM

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
```diff
@@ -3,32 +3,37 @@
 set -o nounset
 set -o errexit
 set -o pipefail
+set -x
```
lslebodn, line 6: Why do we need trace mode in the script ?

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
```diff
@@ -3,32 +3,37 @@
...
+SCORECARD_CONFIG="${SCORECARD_CONFIG:-/tmp/config/scorecard-basic-config.yml}"
 
 # Steps for running the basic operator-sdk scorecard test
 # Expects the standard Prow environment variables to be set and
 # the brew proxy registry credentials to be mounted
 
 NAMESPACE=$(grep "install_namespace:" "${SHARED_DIR}"/oo_deployment_details.yaml | cut -d ':' -f2 | xargs)
 
-OPERATOR_DIR="${ARTIFACT_DIR}/test-operator-basic"
+pushd "${ARTIFACT_DIR}"
```
lslebodn, line 21: There is missing popd counterpart

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
```diff
@@ -3,32 +3,37 @@
...
+OPERATOR_DIR="test-operator-basic"
 
 echo "Starting the basic operator-sdk scorecard test for ${OO_BUNDLE}"
 
 echo "Extracting the operator bundle image into the operator directory"
 mkdir -p "${OPERATOR_DIR}"
 pushd "${OPERATOR_DIR}"
-oc image extract "${OO_BUNDLE}" --confirm -a /var/run/brew-pullsecret/.dockerconfigjson
+oc image extract "${OO_BUNDLE}" --confirm -a "${OPENSHIFT_AUTH}"
+chmod -R go+r ./
```
lslebodn, line 30: please explan why do we need to call chmod here

roles/operator_bundle_scorecard_tests/files/scorecard-basic-config.yml
```diff
@@ -0,0 +1,21 @@
...
+- parallel: true
+  tests:
+  - image: quay.io/operator-framework/scorecard-test:latest
+    entrypoint:
+    - scorecard-test
+    - basic-check-spec
+    labels:
+      suite: basic
+      test: basic-check-spec-test
+  - image: quay.io/operator-framework/scorecard-test:latest
```
samvarankashyap, line 15: isnt there any tagged version. 
should we always use the latest ? 
@14rcole

samvarankashyap (approved):
Looks good to me except for comments.

roles/operator_bundle_scorecard_tests/files/scorecard-basic-config.yml
dirgim: We expect to be able to tag the scorecard images to the v1.8.0 version after we upgrade the operator-sdk to that same version as part of CVP-1887.

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
wheelerlaw: I don't think this could ever hurt. Bash scripts are hard to debug as it is, and having more debug information if something goes wrong could be helpful.

14rcole: Bill had requested that `sed -x` be added so the script was easier to debug.  I don't think there's anything wrong with leaving it there.  If you disagree I can change it, but I'll have to submit a PR to openshift/release as well

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
14rcole: It's on line 31

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
14rcole: When `oc image extract ` extracts the operator bundle, it sets the permissions of the files to `0600`.  We run the scorecard test in a container, and the UUID for that container is not the owner.  Therefore, we need to increase the permissions so the container can access the bundle data.

roles/operator_bundle_scorecard_tests/files/scorecard-basic-config.yml
14rcole: Additionally. this config file is copied from [the operator scorecard container](https://github.com/redhat-operator-ecosystem/operator-scorecard-test-container), which is the container from which scorecard tests are run in DPTP.  Keeping the default configs the same ensures consistency between internal and external test runs.

jsztuka (approved):
LGTM

sonam1412 (approved):
lgtm

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
lslebodn: If somebody want do debug script they can  run `bash -x $name of script` without changing the content.

That file is executed from ansible playbook and it is difficult to read very verbose tracing log in the ansible output unless you run playbook with `-vvv`

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
lslebodn: That is not true. The line 31 is counterpart for the line 28

roles/operator_bundle_scorecard_tests/files/scorecard-basic-config.yml
lslebodn: If we want to have the same file as in redhat-operator-ecosystem/operator-scorecard-test-container
than we should document it somewhere. Maybe in the read.me in the same directory.

roles/operator_bundle_scorecard_tests/files/optional-operators-cvp-common-scorecard-commands.sh
lslebodn: > We run the scorecard test in a container

I assume you meant (DCI) will run the script in k8s/openshift container as unprivileged user ...
Because CVP does not run the script.

And such explanation should be in the script as the comment

roles/operator_bundle_scorecard_tests/files/scorecard-basic-config.yml
lslebodn: And ideally there shouls he comment at the beginning of file as well.

lslebodn (changes requested):
few comments inline

mvalarh (approved):

dirgim (approved):

lslebodn (approved):
LGTM

PR #150, opened 2021-06-04
[CVP-1897] Generate mock prow artifacts when running external Operator Bundle Tests

samvarankashyap:
When operator bundle deployment tests are run through DPTP, two files are generated that the upstream `optional-operators-subscribe` role does not generate.
We will need to generate these files as a part of the `optional-operators-subscribe` role.
The finished.json file should contain the following fields:
timestamp: use the already-existing `start_time` field
passed: boolean true or false
result: Omit. The result field has been deprecated in favor of passed
revision: Omit. The revision field is deprecated
The build-log.txt file is the stdout of the optional-operators-subscribe shell script. We should collect the stdout and save it to the file.
Both files should be saved inside the ARTIFACT_DIR for the role.

### DIFF
--- a/optional-operators-subscribe.yml
+++ b/optional-operators-subscribe.yml
@@ -2,7 +2,7 @@
 - name: "Run optional operators subscribe"
   hosts: localhost
   become: false
-  gather_facts: false
+  gather_facts: true
 
   tasks:
     - name: "Run optional operators subscribe step"
--- a/roles/optional-operators-subscribe/tasks/main.yml
+++ b/roles/optional-operators-subscribe/tasks/main.yml
@@ -9,6 +9,9 @@
       OO_INSTALL_NAMESPACE: "{{ OO_INSTALL_NAMESPACE | default('') }}"
       OO_TARGET_NAMESPACES: "{{ OO_TARGET_NAMESPACES | default('') }}"
       ARTIFACT_DIR: "{{ ARTIFACT_DIR }}"
+      finished_json_contents:
+        passed: false
+        timestamp: "{{ ansible_date_time.iso8601_basic_short }}"
 
   # incase of failure the following task fails
   # we will be ignoring errors to display the results
@@ -31,9 +34,30 @@
     debug:
       msg: "{{ shell_script_output.stdout_lines }}"
 
+  - name: "Set finished_json_contents in case of success"
+    set_fact:
+      finished_json_contents: "{{ finished_json_contents | combine(new_item, recursive=true) }}"
+    vars:
+      new_item: { 'passed':  true }
+    with_dict: "{{ finished_json_contents }}"
+    when: shell_script_output.rc == 0
+
   always:
+  - name: "Store the result of the shell script output to build-log.txt"
+    copy:
+      dest: "{{ ARTIFACT_DIR }}/build-log.txt"
+      content: "{{ shell_script_output.stdout }}"
+      mode: 0644
+
   - name: "Store the result in a JSON file"
     copy:
       dest: "{{ ARTIFACT_DIR }}/optional_operator_subscribe.json"
       content: "{{ shell_script_output }}"
       mode: 0644
+
+  - name: "Store the finished.json"
+    copy:
+      dest: "{{ ARTIFACT_DIR }}/finished.json"
+      content: "{{ finished_json_contents }}"
+      mode: 0644
+


wheelerlaw:
lgtm

roles/optional-operators-subscribe/templates/optional-operators-subscribe-commands.sh
```diff
@@ -152,6 +152,8 @@ target_namespaces: "${OO_TARGET_NAMESPACES}"
 deployment_start_time: "${DEPLOYMENT_START_TIME}"
 EOF
             cp "${ARTIFACT_DIR}/${DEPLOYMENT_ART}" "${SHARED_DIR}/${DEPLOYMENT_ART}"
+	    # Generate finished.json at end of the run
+	    printf '{"timestamp":"%s","passed": true }' "${DEPLOYMENT_START_TIME}" > "${ARTIFACT_DIR}/finished.json"
```
14rcole, line 156: Nit: fix indentation here

14rcole (approved):
One small comment, otherwise LGTM

roles/optional-operators-subscribe/templates/optional-operators-subscribe-commands.sh
```diff
@@ -205,4 +207,5 @@ else
     echo "Dumping all ClusterServiceVersions in namespace $OO_INSTALL_NAMESPACE to $CSV_ART"
     oc get -n "$OO_INSTALL_NAMESPACE" csv -o yaml >"$CSV_ART"
 fi
+printf '{"timestamp":"%s","passed": false }' "${DEPLOYMENT_START_TIME}" > "${ARTIFACT_DIR}/finished.json"
```
14rcole, line 210: Actually, I have to rescind my LGTM.  Because this script is shared with DPTP (we just copy the file from `openshift/release`), we can't make changes here without submitting them in that repo as well.  Can you move the generation of this file to the ansible playbook. The `collect_modified_images` role has an example of how to get the timestamps if you don't know how.

14rcole (changes requested):

roles/optional-operators-subscribe/templates/optional-operators-subscribe-commands.sh
samvarankashyap: Will fix it.

roles/optional-operators-subscribe/templates/optional-operators-subscribe-commands.sh
samvarankashyap: Got it. Thanks

samvarankashyap: @14rcole How specific are you about the timestamp. 
Would the timestamp mean about the deployment start time ( which is a dptp delpoyment start time am not sure about how to get it ) ? 
the optional operator subscribe step start time ? 
If the later  i can add a variable to fetch the timestamp within this playbook ...
could you clarify

14rcole: If by the second you mean the start time for the shell script then yes, do that

The script itself is the deployment.  When this playbook is run, we won't actually be calling out to DPTP

jsztuka (approved):
Lgtm

roles/optional-operators-subscribe/templates/optional-operators-subscribe-commands.sh
samvarankashyap: I have made necessary changes please re-review 
@14rcole

samvarankashyap:
@14rcole bump

optional-operators-subscribe.yml
```diff
@@ -2,7 +2,7 @@
 - name: "Run optional operators subscribe"
   hosts: localhost
   become: false
-  gather_facts: false
+  gather_facts: true
```
dirgim, line 5: Just a quick point - depending on the DCI's environment, gathering facts might fail for them. 
@14rcole do you think if this is a legitimate worry on my part?

samvarankashyap: @dirgim Gather facts should work with any env since it takes variables related to the system. 
Why do you think it doesnt work ?

samvarankashyap:
@14rcole bump

optional-operators-subscribe.yml
dirgim: I've seen it fail in some weird environments in the past. But I think it should work fine in DCI environments since they use Ansible for other tasks too, closing this discussion.

dirgim:
lgtm

dirgim (approved):

14rcole:
LGTM

14rcole (approved):

PR #152, opened 2021-06-14
[CVP-2052] Fix failures when running the midstream-cvp image within an OpenShift pod

samvarankashyap:
the problem arise from the fact that the container runs within an OpenShift pod in more restrictive manner. For e.g., it doesn’t run as root.

For resolving this issue, we have discussed two alternative solutions:

Using product-ci-cd's entrypoint script:https://code.engineering.redhat.com/gerrit/plugins/gitiles/product-ci-cd/+/refs/heads/master/container/midstream/entrypoint.sh and also adding the following to the Dockerfile:

### DIFF
--- a/Dockerfiles/midstream/Dockerfile
+++ b/Dockerfiles/midstream/Dockerfile
@@ -3,6 +3,7 @@ WORKDIR /project
 ARG OPERATOR_SDK_VERSION=v1.4.0
 ARG OPERATOR_TEST_PLAYBOOKS_TAG=v1.0.11
 ARG UMOCI_VERSION=v0.4.5
+ADD ./entrypoint.sh /entrypoint.sh
 RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(arch) ;; esac);\
     export OS=$(uname | awk '{print tolower($0)}');\
     export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/;\
@@ -11,6 +12,7 @@ RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm6
     curl -fL -o /usr/local/bin/umoci https://github.com/opencontainers/umoci/releases/download/${UMOCI_VERSION}/umoci.amd64 && \
     chmod a+x /usr/local/bin/umoci && \
     mkdir /project/output && \
+    chmod g+w /etc/passwd && \
     dnf install --setopt=install_weak_deps=False -y git-core ansible skopeo && \
     dnf clean all
 ADD ./run_tests.py /run_tests.py
--- a/Dockerfiles/midstream/Dockerfile
+++ b/Dockerfiles/midstream/Dockerfile
@@ -3,7 +3,7 @@ WORKDIR /project
 ARG OPERATOR_SDK_VERSION=v1.4.0
 ARG OPERATOR_TEST_PLAYBOOKS_TAG=v1.0.11
 ARG UMOCI_VERSION=v0.4.5
-ADD ./entrypoint.sh /entrypoint.sh
+ADD ./fix_etc_passwd.sh /usr/bin/fix_etc_passwd.sh
 RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(arch) ;; esac);\
     export OS=$(uname | awk '{print tolower($0)}');\
     export OPERATOR_SDK_DL_URL=https://github.com/operator-framework/operator-sdk/releases/download/$OPERATOR_SDK_VERSION/;\
--- a/Dockerfiles/midstream/entrypoint.sh
+++ b/Dockerfiles/midstream/entrypoint.sh
@@ -1,0 +1,35 @@
+#!/bin/bash -e
+main() {
+    local username="${SSH_AUTH_USER:-default}"
+    local home
+    home=$(mk_home "$username")
+    setup_passwd "$username" "$home"
+    exec sleep inf
+}
+mk_home() {
+    local username="${1:?}"
+    [[ -d /workDir ]] && [[ -w /workDir ]] && \
+        mkdir -p "/workDir/home/$username" && \
+        echo "/workDir/home/$username" && \
+        return 0
+    mktemp --tmpdir -d "${username}.XXXXXX"
+}
+setup_passwd() {
+    local username="${1:?}"
+    local home="${2:?}"
+    local uid="$3"
+    local gid="${4:-0}"
+    local passwd=/etc/passwd
+    ! whoami 2>/dev/null || return 0
+    if ! [[ -w "$passwd" ]]; then
+        echo "There is no permission to add user $username to $passwd"
+        return 0
+    fi
+    [[ "$uid" ]] || uid="$(id -u)"
+    echo \
+        "${username}:x:${uid}:${gid}:Default Application User:${home}:/sbin/nologin" \
+        >> /etc/passwd
+}
+if [[ "${BASH_SOURCE[0]}" == "$0" ]]; then
+    main "$@"
+fi
--- a/Dockerfiles/midstream/fix_etc_passwd.sh
+++ b/Dockerfiles/midstream/fix_etc_passwd.sh
@@ -4,26 +4,23 @@ main() {
     local home
     home=$(mk_home "$username")
     setup_passwd "$username" "$home"
-    exec sleep inf
 }
 mk_home() {
     local username="${1:?}"
-    [[ -d /workDir ]] && [[ -w /workDir ]] && \
-        mkdir -p "/workDir/home/$username" && \
-        echo "/workDir/home/$username" && \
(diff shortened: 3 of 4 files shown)


Dockerfiles/midstream/Dockerfile
```diff
@@ -13,6 +13,8 @@ RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm6
     mkdir /project/output && \
     dnf install --setopt=install_weak_deps=False -y git-core ansible skopeo && \
     dnf clean all
+ADD ./entrypoint.sh /entrypoint.sh
```
dirgim, line 16: Can you move this `ADD` instruction to be above line 6 and then incorporate the `RUN chmod g+x /entrypoint.sh && chmod g+w /etc/passwd` commands into the `RUN` instruction at line 6?

samvarankashyap: Sure I can do that.

Dockerfiles/midstream/Dockerfile
```diff
@@ -11,6 +12,8 @@ RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm6
     curl -fL -o /usr/local/bin/umoci https://github.com/opencontainers/umoci/releases/download/${UMOCI_VERSION}/umoci.amd64 && \
     chmod a+x /usr/local/bin/umoci && \
     mkdir /project/output && \
+    chmod g+x /entrypoint.sh && \
```
lslebodn, line 15: You do not need such step if file bits are already properly set in git repository

Dockerfiles/midstream/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+    [[ -d /workDir ]] && [[ -w /workDir ]] && \
+        mkdir -p "/workDir/home/$username" && \
+        echo "/workDir/home/$username" && \
+        return 0
+    mktemp --tmpdir -d "${username}.XXXXXX"
+}
+setup_passwd() {
+    local username="${1:?}"
+    local home="${2:?}"
+    local uid="$3"
```
lslebodn, line 20: Looks like we do not pass UID as parameter. 
We can simplify the code and directly use `uid="$(id -u)"`

Dockerfiles/midstream/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+    local gid="${4:-0}"
+    local passwd=/etc/passwd
+    ! whoami 2>/dev/null || return 0
```
lslebodn, line 23: We are interested in entry in /etc/passwd. Might be better to check it directly.
```
getent passwd $uid
```

Dockerfiles/midstream/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+    local home
+    home=$(mk_home "$username")
+    setup_passwd "$username" "$home"
+    exec sleep inf
+}
+mk_home() {
+    local username="${1:?}"
+    [[ -d /workDir ]] && [[ -w /workDir ]] && \
+        mkdir -p "/workDir/home/$username" && \
+        echo "/workDir/home/$username" && \
```
lslebodn, line 13: I would rather create it in `/tmp` than in `/workDir`

lslebodn:
@samvarankashyap the script `entrypoint.sh` is written in a way it can be executed always
even for case where entry already exists in /etc/passwd.

But It is assumed users of the image will run it themselves. IMHO we can run it ourselves from the `run_tests.py`
We we can run even tests in gitlab action with passing uid to podman
e.g. `podman run --user 314159:0`

Dockerfiles/midstream/entrypoint.sh
samvarankashyap: Updated.

Dockerfiles/midstream/Dockerfile
samvarankashyap: updated

Dockerfiles/midstream/entrypoint.sh
samvarankashyap: How would it effect the run for changing to /tmp/ ?
I understand it might be the permissions. 
How would this help ?

samvarankashyap:
@lslebodn Do you mean invoking the entrypoint.sh from run_tests.py 
with appropriate arguments ? 
note: I have borrowed the script from the cpaas with no changes except for calling run_tests.py.

Dockerfiles/midstream/entrypoint.sh
samvarankashyap: updating meanwhile in subsequent commits

lslebodn: `workDir` is a non-standard directly and we should not rely on it.

lslebodn:
> Do you mean invoking the entrypoint.sh from run_tests.py
Yes

Dockerfiles/midstream/entrypoint.sh
samvarankashyap: Resolved

samvarankashyap:
@lslebodn 
I tried switiching the logic to run shell script inside the python script run_tests.py 
for reason unknown to me it caused a hungup in both local run and github actions , 
ref: https://github.com/redhat-operator-ecosystem/operator-test-playbooks/pull/152/commits/490b5b426d8f73717c35b3814b10564e93d1427b 

However, When i switched it back to running python script inside the run_tests.py 
it worked absolutely fine. 
ref: https://github.com/redhat-operator-ecosystem/operator-test-playbooks/pull/152/commits/13c100e3895baf5aa312414a8a0fd18b2a423c0d 
I have reverted few changes. However, I made changes to the shell script as requested. 
Please review. 
I will rebase and squash once its re-reviewed 
Thanks. 
Tagging @dirgim @jsztuka for review

Dockerfiles/midstream/entrypoint.sh
```diff
@@ -0,0 +1,34 @@
...
+    mkdir -p "/tmp/home/$username" && \
+    echo "/tmp/home/$username" && \
+    return 0
+    mktemp --tmpdir -d "${username}.XXXXXX"
+}
+setup_passwd() {
+    local username="${1:?}"
+    local home="${2:?}"
+    local uid="$(id -u)"
+    local gid="${4:-0}"
```
dirgim, line 20: I think it makes sense to change the `gid` to be the 3rd parameter as `uid` isn't tied to a parameter anymore.

jsztuka (approved):
Looks good, just a small nit that Kruno suggested.

dirgim (approved):
LGTM

lslebodn:
> for reason unknown to me it caused a hungup in both local run and github actions

@samvarankashyap because the cpaas sript was written in a way that would never finish. That was their use-case

Removing `sleep inf` will help
```
diff --git a/Dockerfiles/midstream/fix_etc_passwd.sh b/Dockerfiles/midstream/fix_etc_passwd.sh
index 44b768a..9562285 100644
--- a/Dockerfiles/midstream/fix_etc_passwd.sh
+++ b/Dockerfiles/midstream/fix_etc_passwd.sh
@@ -4,7 +4,6 @@ main() {
     local home
     home=$(mk_home "$username")
     setup_passwd "$username" "$home"
-    exec sleep inf
 }
 mk_home() {
     local username="${1:?}"
```

lslebodn:
BTW do not add the script `fix_etc_passwd.sh` into root.
Install it to standard location /usr/bin or /us/local/bin.
It will simplify execution as well. because you will be able to get rid of logic with `fix_etc_passwd_path`
keep it simple and functional :-)

Dockerfiles/midstream/run_tests.py
```diff
@@ -88,6 +88,9 @@ def test_for_extract_and_validate_bundle_image(self):
 
 if __name__ == '__main__':
     resultcodes = [0, 50, 70, 102, 103]
+    # run fix_etc_passwd.sh to setup random user generated in openshift
+    fix_etc_passwd_path = path.dirname(path.realpath(__file__))
```
lslebodn, line 92: unused variable
Otherwise lgtm

samvarankashyap: done

Dockerfiles/midstream/fix_etc_passwd.sh
```diff
@@ -0,0 +1,32 @@
...
+    return 0
+    mktemp --tmpdir -d "${username}.XXXXXX"
+}
+setup_passwd() {
+    local username="${1:?}"
+    local home="${2:?}"
+    local uid="$(id -u)"
+    local gid="${4:-0}"
+    ! whoami 2>/dev/null || return 0
+    if ! [[ $(getent passwd $uid) ]]; then
```
lslebodn, line 21: @samvarankashyap I am not sure whether I wrote comment to different line or it was changed on different line.
It was
```
     ! whoami 2>/dev/null || return 0
     if ! [[ -w "$passwd" ]]; then
         echo "There is no permission to add user $username to $passwd"
        return 0
```
And it should be
```
     ! getent passwd $uid 2>/dev/null || return 0
     if ! [[ -w "$passwd" ]]; then
         echo "There is no permission to add user $username to $passwd"
        return 0

```

Eran found the issue

PR #160, opened 2021-06-24
[CVP-2109] Add playbooks and update calls to multiple processes

samvarankashyap:
$subject

PR #196, opened 2021-07-15
[CVP-2083] Updated upstream cvp-midstream-image endpoint

samvarankashyap:
Update the midstream cvp image endpoint,

### DIFF
--- a/.github/workflows/build_and_push_image_on_tag.yml
+++ b/.github/workflows/build_and_push_image_on_tag.yml
@@ -36,4 +36,4 @@ jobs:
           cd Dockerfiles/midstream/
           podman build -t midstream_image:$latestTag -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=$latestTag
           podman login -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io
-          podman push midstream_image:$latestTag quay.io/samvarankashyap/cvp-midstream-image:$latestTag
+          podman push midstream_image:$latestTag quay.io/cvpops/cvp-midstream-image:$latestTag
--- a/.github/workflows/build_and_push_image_on_tag.yml
+++ b/.github/workflows/build_and_push_image_on_tag.yml
@@ -33,7 +33,6 @@ jobs:
           git checkout $latestTag
           echo "Latest tag is $latestTag"
           podman --version
-          cd Dockerfiles/midstream/
-          podman build -t midstream_image:$latestTag -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=$latestTag
+          podman build -t midstream_image:$latestTag -f Dockerfiles/midstream/Dockerfile .
           podman login -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io
           podman push midstream_image:$latestTag quay.io/cvpops/cvp-midstream-image:$latestTag
--- a/.github/workflows/build_and_push_latest_image.yml
+++ b/.github/workflows/build_and_push_latest_image.yml
@@ -35,4 +35,4 @@ jobs:
           cd Dockerfiles/midstream/
           podman build -t midstream_image:latest -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=master
           podman login -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io
-          podman push midstream_image:latest quay.io/samvarankashyap/cvp-midstream-image
+          podman push midstream_image:latest quay.io/cvpops/cvp-midstream-image:latest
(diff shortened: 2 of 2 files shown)


samvarankashyap:
This should be merged after #160

.github/workflows/build_and_push_latest_image.yml
```diff
@@ -35,4 +35,4 @@ jobs:
           cd Dockerfiles/midstream/
           podman build -t midstream_image:latest -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=master
           podman login -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io
-          podman push midstream_image:latest quay.io/samvarankashyap/cvp-midstream-image
+          podman push midstream_image:$latestTag quay.io/cvpops/cvp-midstream-image:$latestTag
```
lslebodn, line 38: Where is the `$latestTag` defined ?

samvarankashyap: I missed a line there. 
I fixed it.

lslebodn: IIUC the goal of the github workflow is to push latest master after merging.
So we should not overwrite any tag. So IMHO the usage of `latest` was good

14rcole (approved):
LGTM

.github/workflows/build_and_push_latest_image.yml
samvarankashyap: @lslebodn 
```
on:
  # Triggers the workflow on tag events events but only for the master branch
  push:
    tags:
      - '*'
  # Allows you to run this workflow manually from the Actions tab
```
according to the syntax, this workflow only gets triggered in case of creation of new tags. 
Thus only whenever a new tag is created the image is built and pushed . 
hope this answers. We will not be overwriting any tags. 
we will only create a new tag. or push to the tags which we do not do in our case.

lslebodn: @samvarankashyap  that triiger is in the different file.
I can see 
```
# This is a basic workflow to help you get started with Actions
name: "Build and push to quay on every commit"
# Controls when the action will run.
on:
  # Triggers the workflow on push or pull request events but only for the master branch
  push:
    branches: [ master ]
```
in this file therefore usage of `latest` make more sense to me

samvarankashyap: Yes. True. I think i was talking about the other file. 
I will change it ASAP

lslebodn (approved):
+1

samvarankashyap:
Should be merged after #160

.github/workflows/build_and_push_image_on_tag.yml
```diff
@@ -33,7 +33,6 @@ jobs:
           git checkout $latestTag
           echo "Latest tag is $latestTag"
           podman --version
-          cd Dockerfiles/midstream/
-          podman build -t midstream_image:$latestTag -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=$latestTag
+          podman build -t midstream_image:$latestTag -f Dockerfiles/midstream/Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=$latestTag
```
lslebodn, line 36: The tag `OPERATOR_TEST_PLAYBOOKS_TAG` is not used in any dockerfile

We should rely on default value in Dockerfile for `OPERATOR_SDK_VERSION`
Otherwise people will need to update the same number on more places.
And people tend to forget :-) We should just remove it from github actions

.github/workflows/build_and_push_latest_image.yml
```diff
@@ -32,7 +32,6 @@ jobs:
           QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
         run: |
           podman --version
-          cd Dockerfiles/midstream/
-          podman build -t midstream_image:latest -f Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.4.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=master
+          podman build -t midstream_image:latest -f Dockerfiles/midstream/Dockerfile --build-arg OPERATOR_SDK_VERSION=v1.9.0 --build-arg OPERATOR_TEST_PLAYBOOKS_TAG=master
```
lslebodn, line 35: The tag `OPERATOR_TEST_PLAYBOOKS_TAG` is not used in any Dockerfile

We should rely on default value in Dockerfile for `OPERATOR_SDK_VERSION`
Otherwise people will need to update the same number on more places.
And people tend to forget :-) We should just remove it from github actions

lslebodn (changes requested):

.github/workflows/build_and_push_image_on_tag.yml
lslebodn: @samvarankashyap bump

.github/workflows/build_and_push_latest_image.yml
lslebodn: @samvarankashyap bump

samvarankashyap: Updated.

lslebodn (approved):
LGTM

dirgim (approved):
LGTM

PR #231, opened 2021-09-07
[CVP-2252] Add checks for deprecated APIs to midstream pipeline

14rcole:
The OCP 4.9 operator bundles will have to be tested to ensure that they do not use any deprecated apis.  This PR modifieds the validate_operator_bundle role to determine if the bundle being tested supports v4.9 and, if it does, add's a flag to the bundle validation tests to check for the deprecated APIs.

### DIFF
--- a/Dockerfiles/midstream/unit_tests.py
+++ b/Dockerfiles/midstream/unit_tests.py
@@ -31,7 +31,7 @@ class Testing(unittest.TestCase):
 
     # Running the container with correctly set environment variables, default channel missing, all tasks should pass
     def test_positive_missing_default_channel(self):
-        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:missing-default-channel-v1"
+        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:missing-default-channel-v2"
         result = subprocess.run(self.exec_cmd,
                                 shell=True,
                                 env=self.env,
@@ -80,10 +80,27 @@ class Testing(unittest.TestCase):
                       "Result code not found in %s" % message)
         self.assertEqual(102, result.returncode)
         self.assertTrue(os.path.exists(OUTPUT_DIR+".errormessage"))
-    
+
+    # Run with a test-operator which fails the deprecated image check in the
+    # bundle image validation job
+    def test_default_negative(self):
+        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:test-default-negative-v1"
+        result = subprocess.run(self.exec_cmd,
+                                shell=True,
+                                env=self.env,
+                                cwd=OUTPUT_DIR)
+        self.assertEqual(70, result.returncode)
+        # check if the .errormessage file generated is empty
+        self.assertTrue(os.stat("/tmp/.errormessage").st_size != 0)
+        # check if the .errormessage exists since its a logger file
+        self.assertTrue(os.path.exists(OUTPUT_DIR+".errormessage"))
+        message = self.get_error_message_content()
+        self.assertIn("this bundle is using APIs which were deprecated and removed in v1.22", message,
+                      "Deprecated APIs error not found in '%s'" % message)
+
     # Run with a test-operator which passes the bundle image validation job
     def test_default_positive(self):
-        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:v1.0-16"
+        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:test-default-positive-v1"
         result = subprocess.run(self.exec_cmd,
                                 shell=True,
                                 env=self.env,
--- a/roles/validate_operator_bundle/tasks/main.yml
+++ b/roles/validate_operator_bundle/tasks/main.yml
@@ -28,10 +28,69 @@
     bundle_validate: "{{ operator_work_dir }}"
   when: not run_upstream|bool
 
-- name: "Set the operatorhub checks if enabled"
+- name: "list work_dir"
+  shell: "ls {{ work_dir }}"
+
+- name: "Check whether bundle supports v4.9 or higher"
+  block:
+    - name: Get skopeo inspect json
+      shell: "cat {{ work_dir }}/bundle-skopeo-inspect.json"
+      register: bundle_skopeo_inspect_json
+
+    - name: Convert json to map
+      set_fact:
+        bundle_skopeo_inspect_map: "{{ bundle_skopeo_inspect_json.stdout | from_json }}"
+
+    - name: Determine ocp version annotation
+      set_fact:
+        ocp_version_annotation: "{{ bundle_skopeo_inspect_map['Labels']['com.redhat.openshift.versions'] }}"
+
+    - name: "Determine OCP version range"
+      uri:
+        url: "https://catalog.redhat.com/api/containers/v1/operators/indices?sort_by=data.ocp_version&ocp_versions_range={{ ocp_version_annotation }}&organization=redhat-operators"
+        method: GET
+        headers:
+          Accept: "application/json"
+        status_code: 200
+        return_content: true
+      register: supported_indices_json
+      retries: 10
+      ignore_errors: true
+
+    - name: "Fail if Pyxis response is not 200"
+      fail:
+        msg: "Error collecting OCP version range from Pyxis.  Error message: {{ supported_indices_json.content }}"
+      when: supported_indices_json.status != 200
+
+    - name: "Convert output to map"
+      set_fact:
+        supported_indices: "{{ supported_indices_json.content | from_json }}"
+
+    - name: debug
+      debug:
+        msg: "Indices: {{ supported_indices['data'] }}"
+
+    - name: "Set support_v4_9 to false if there are no supported versions"
+      set_fact:
+        support_v4_9: false
+      when: supported_indices['data']|length==0
+
+    - name: "Parse values returned from Pyxis"
+      set_fact:
+        highest_index: "{{ supported_indices['data'] | last }}"
+      when: supported_indices['data']|length>0
+
+    - name: "Determine if v4.9 is supported"
+      set_fact:
+        support_v4_9: "{{ highest_index['ocp_version'] is version('4.9', '>=') }}"
+      when:
+        - highest_index is defined
+        - highest_index['ocp_version']|length>0
+
+- name: "Set the operatorhub checks if bundle supports v4.9 or greater"
   set_fact:
-    select_optional_tests: "--select-optional name=operatorhub"
-  when: run_operatorhub_checks|bool
+    select_optional_tests: "--select-optional name=operatorhub --optional-values=k8s-version=1.22"
+  when: support_v4_9
 
 - name: "Validate the operator bundle manifest and metadata with operator-sdk bundle validate"
   shell: "{{ operator_sdk_bin_path }} bundle validate {{ select_optional_tests }} --verbose {{ bundle_validate }} 2>&1"


roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -26,12 +26,51 @@
 - name: "Set bundle_validate variable"
   set_fact:
     bundle_validate: "{{ operator_work_dir }}"
-  when: not run_upstream|bool
+    when: not run_upstream|bool
```
dirgim, line 29: I think this should stay indented as before

14rcole: Thanks, vim often messes up the auto-indentation for ansible

dirgim:
The images used in unit tests have deprecated APIs, so are now [failing the tests](https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/pull/231/checks?check_run_id=3535438469). We'll need to make sure that we both cover the case where the image fails in the unit tests as well as provide an updated image that has APIs that pass the new validation.

dirgim:
For the positive case we can use the image at `quay.io/cvpops/test-operator:test-default-positive-v1` instead the existing outdated one here: https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/blob/master/Dockerfiles/midstream/unit_tests.py#L86

Dockerfiles/midstream/unit_tests.py
```diff
@@ -31,7 +31,7 @@ def get_error_message_content(output_file=".errormessage"):
 
     # Running the container with correctly set environment variables, default channel missing, all tasks should pass
     def test_positive_missing_default_channel(self):
-        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:missing-default-channel-v1"
+        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:test-default-positive-v1"
```
14rcole, line 34: @dirgim is the image you gave me missing the default channel, or will I need to build a different image for this test

dirgim: It's not missing a default channel, but I built one that does (while also passing the deprecated API check) - quay.io/cvpops/test-operator:missing-default-channel-v2

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -28,10 +28,49 @@
...
+
+    - name: "Query Pyxis"
+      uri:
+        url: "https://catalog.redhat.com/api/containers/v1/operators/indices?sort_by=data.ocp_version&ocp_versions_range={{ ocp_version_annotation }}&organization=redhat-operators"
+        method: GET
+        headers:
+          Accept: "application/json"
+        status_code: 200
+        return_content: true
+      register: supported_indices_json
```
lslebodn, line 56: Please use few retries here to avoid random network issues

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -28,10 +28,49 @@
...
+
+    - name: Convert json to map
+      set_fact:
+        bundle_skopeo_inspect_map: "{{ bundle_skopeo_inspect_json.stdout | from_json }}"
+
+    - name: Determine ocp version annotation
+      set_fact:
+        ocp_version_annotation: "{{ bundle_skopeo_inspect_map['Labels']['com.redhat.openshift.versions'] }}"
+
+    - name: "Query Pyxis"
```
lslebodn, line 48: Parse ocp version range is bit more description.
Using pyxis for that is jsut implementation detail

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -28,10 +28,49 @@
...
+      uri:
+        url: "https://catalog.redhat.com/api/containers/v1/operators/indices?sort_by=data.ocp_version&ocp_versions_range={{ ocp_version_annotation }}&organization=redhat-operators"
+        method: GET
+        headers:
+          Accept: "application/json"
+        status_code: 200
```
lslebodn, line 54: We need to somehow handle and do nice error report for invalid syntax
e.g. 
```
[alcik@localhost][/tmp/data]$ curl "https://catalog.redhat.com/api/containers/v1/operators/indices?sort_by=data.ocp_version&ocp_versions_range=v4.9,v4.10&organization=redhat-operators" | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   245  100   245    0     0    413      0 --:--:-- --:--:-- --:--:--   412
{
  "detail": "CSV filter is not supported for versions other than: (4.5,4.6,4.7,4.8) in any order. Unsupported version(s): (4.9,4.10)",
  "status": 400,
  "title": "Bad Request",
  "type": "about:blank",
  "trace_id": "0xffce4c92d42874d98e20ba0960cc59f4"
}
```

+ add code coverage for that

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -28,10 +28,49 @@
...
+        return_content: true
+      register: supported_indices_json
+
+    - name: "Convert output to map"
+      set_fact:
+        supported_indices: "{{ supported_indices_json.content | from_json }}"
+
+    - name: "Parse values returned from Pyxis"
+      set_fact:
+        highest_index: "{{ supported_indices['data'] | last }}"
```
lslebodn, line 64: Will this work empty data? e.g. non-existing ocp range `=4.1`
```
sh$ curl "https://catalog.redhat.com/api/containers/v1/operators/indices?sort_by=data.ocp_version&ocp_versions_range==v4.1&organization=redhat-operators" | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    64  100    64    0     0    172      0 --:--:-- --:--:-- --:--:--   172
{
  "data": [],
  "page": 0,
  "page_size": 100,
  "total": 0
}
```

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -28,10 +28,49 @@
...
+
+    - name: "Determine if v4.9 is supported"
+      set_fact:
+        support_v4_9: "{{ highest_index['ocp_version'] is version('4.9', '>=') }}"
```
lslebodn, line 68: Code freeze for OCP v4.9 already happen so we can use stricter check
and `support_v4_9` should be true when v4.9 is in the range.

older version was weaker "only v4.9" was supported

Dockerfiles/midstream/unit_tests.py
```diff
@@ -80,8 +80,9 @@ def test_negative_image_to_test_not_set(self):
                       "Result code not found in %s" % message)
         self.assertEqual(102, result.returncode)
         self.assertTrue(os.path.exists(OUTPUT_DIR+".errormessage"))
-    
-    # Run with a test-operator which passes the bundle image validation job
+
+    # Run with a test-operator which fails the deprecated image check in the
+    # bundle image validation job
```
lslebodn, line 85: If we change the intention of test-case with the comment then we should check new "expected error" message in some output.

Dockerfiles/midstream/unit_tests.py
```diff
@@ -94,6 +95,19 @@ def test_default_positive(self):
...
+    # Run with a test-operator which passes the bundle image validation job
+    def test_default_positive(self):
+        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:test-default-positive-v1"
+        result = subprocess.run(self.exec_cmd,
+                                shell=True,
+                                env=self.env,
+                                cwd=OUTPUT_DIR)
+        self.assertEqual(0, result.returncode)
+        # check if the .errormessage file generated is empty
+        self.assertTrue(os.stat("/tmp/.errormessage").st_size == 0)
```
lslebodn, line 107: Do not use hardcoded `/tmp` other palaces used `OUTPUT_DIR`

lslebodn:
Cannot see negative test-case with some expected failed output for v4.9 deprecated api.

roles/validate_operator_bundle/tasks/main.yml
14rcole: This check covers the stricter version.  It lists all of the supported OCP versions and if the highest version is 4.9 or greater, it adds the check

14rcole:
> Cannot see negative test-case with some expected failed output for v4.9 deprecated api.

That's because the test cases are covered in the downstream repo.  I'm not sure if I should be linking from an internal repository from GitHub, but the test you're looking for is in `OperatorUtilsTest.groovy` on line 53

roles/validate_operator_bundle/tasks/main.yml
14rcole: It will now.  Fixed.

dirgim:
@J0zi @mvalarh can you take a look at the Operator test playbook upstream (otpu) tests / Index from production (pull_request) failure and help us figure out how to get it fixed?

mvalarh:
@14rcole and @dirgim this test can fail now because master branch doesn't have latest changes in `upstream-community`. We can ignore our tests for now in pull requests to master. I will make PR to fix it

mvalarh:
@dirgim please merge this PR (https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/pull/241) and then rerun tests. Thanks

Dockerfiles/midstream/unit_tests.py
```diff
@@ -80,8 +80,9 @@ def test_negative_image_to_test_not_set(self):
                       "Result code not found in %s" % message)
         self.assertEqual(102, result.returncode)
         self.assertTrue(os.path.exists(OUTPUT_DIR+".errormessage"))
-    
-    # Run with a test-operator which passes the bundle image validation job
+
+    # Run with a test-operator which fails the deprecated image check in the
+    # bundle image validation job
     def test_default_positive(self):
```
dirgim, line 86: Can you change the name of this test and adjust the assertions on line 92? Currently this unit test is not executed because it has the same name as the one below it.

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -28,10 +28,70 @@
...
+        bundle_skopeo_inspect_map: "{{ bundle_skopeo_inspect_json.stdout | from_json }}"
+
+    - name: Determine ocp version annotation
+      set_fact:
+        ocp_version_annotation: "{{ bundle_skopeo_inspect_map['Labels']['com.redhat.openshift.versions'] }}"
+
+    - name: "Determine OCP version range"
+      uri:
+        #url: "https://catalog.redhat.com/api/containers/v1/operators/indices?sort_by=data.ocp_version&ocp_versions_range={{ ocp_version_annotation }}&organization=redhat-operators"
+        url: "https://catalog.redhat.com/api/containers/v1/operators/indices?sort_by=data.ocp_version&ocp_versions_range=4.1-4.3&organization=redhat-operators"
```
dirgim, line 51: Can you remove this line and uncomment the one above it?

Dockerfiles/midstream/unit_tests.py
```diff
@@ -80,10 +80,24 @@ def test_negative_image_to_test_not_set(self):
                       "Result code not found in %s" % message)
         self.assertEqual(102, result.returncode)
         self.assertTrue(os.path.exists(OUTPUT_DIR+".errormessage"))
-    
+
+    # Run with a test-operator which fails the deprecated image check in the
+    # bundle image validation job
+    def test_default_negative(self):
+        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:v1.0-16"
```
dirgim, line 87: This bundle seems to have an invalid ocp version support label `v4.5-test` which results in an empty list returned from Pyxis. I've made an updated version of this bundle image with the correct `v4.5` label and pushed it to `quay.io/cvpops/test-operator:test-default-negative-v1`

14rcole:
Every one of these tests is failing in the same way:
```
TASK [parse_operator_bundle : Result of failed task] *****************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": {
        "changed": false,
        "failed": true,
        "msg": "Required operator bundle image label(s) not found:  com.redhat.openshift.versions"
    }
}

TASK [parse_operator_bundle : Result of failure] *********************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Ansible playbook while preparing the operator metadata failed"}

```

This step occurs during the operator bundle extraction stage.  Three of the four failing tests came from images that were changed in this MR, so one of the labels could have been set incorrectly.  The fourth test, `test_negative_image_bundle_validation`, should not be failing.  It is unmodified in this MR.  I'm continuing to investigate what could be wrong

14rcole:
It's also worth noting that when I run `skopeo inspect <image>` in the images used in each test, they all have the `com.redhat.openshift.versions` label set.  Additionally (this is probably irrellevant) most images have the label set to "v4.5"

dirgim (approved):
LGTM

jsztuka (approved):
LGTM

Dockerfiles/midstream/unit_tests.py
```diff
@@ -80,10 +80,27 @@ def test_negative_image_to_test_not_set(self):
...
+    # bundle image validation job
+    def test_default_negative(self):
+        self.env["IMAGE_TO_TEST"] = "quay.io/cvpops/test-operator:test-default-negative-v1"
+        result = subprocess.run(self.exec_cmd,
+                                shell=True,
+                                env=self.env,
+                                cwd=OUTPUT_DIR)
+        self.assertEqual(70, result.returncode)
+        # check if the .errormessage file generated is empty
+        self.assertTrue(os.stat("/tmp/.errormessage").st_size != 0)
```
lslebodn, line 94: The `tmp` must not be hardcoded

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -28,10 +28,69 @@
...
+          Accept: "application/json"
+        status_code: 200
+        return_content: true
+      register: supported_indices_json
+      retries: 10
+      ignore_errors: true
+
+    - name: "Fail if Pyxis response is not 200"
+      fail:
+        msg: "Error collecting OCP version range from Pyxis.  Error message: {{ supported_indices_json.content }}"
```
lslebodn, line 62: We should log also http status code

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -28,10 +28,69 @@
...
+        supported_indices: "{{ supported_indices_json.content | from_json }}"
+
+    - name: debug
+      debug:
+        msg: "Indices: {{ supported_indices['data'] }}"
+
+    - name: "Set support_v4_9 to false if there are no supported versions"
+      set_fact:
+        support_v4_9: false
+      when: supported_indices['data']|length==0
```
lslebodn, line 76: If there are not data from pyxis than we should fail.

lslebodn (commented):
> That's because the test cases are covered in the downstream repo.

We wrote a new feature it has to have code coverage in upstream. We must not rely on any downstream test.
Especially if it is possible to test it in upstream

PR #244, opened 2021-09-16
fix: updates URLs in test.sh after repo move

bartoszmajsak:
After moving the repo `test.sh` is still referencing old URLs thus failing when running without overwritten env vars.


bartoszmajsak:
I don't know why `CVP/shared-lib-tests` is failing as I don't have access to details. Can someone shed some light?

lslebodn:
@J0zi  Is upstream directory used from the master branch?
Maybe MR should be created to different branch

PR #246, opened 2021-09-16
Updates to tests for deprecated APIs check

14rcole:
Per [this comment](https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/pull/231#pullrequestreview-756159952) I added another test for the validate_operator_bundle role and fixed a couple of other nits

### DIFF
--- a/.github/workflows/build_and_test_upstream_ci_image.yml
+++ b/.github/workflows/build_and_test_upstream_ci_image.yml
@@ -32,3 +32,15 @@ jobs:
         run: |
           mkdir test_operator_work_dir output_logs
           podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_success_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
+      
+      - name: "Run the bundle validation test for a bundle that supports a very old OCP version --> negative test"
+        continue-on-error: false
+        run: |
+          mkdir test_operator_work_dir output_logs
+          podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_success_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
+      
+      - name: "Run the bundle validation test for a bundle with a syntactically incorrect OCP version range --> negative test"
+        continue-on-error: false
+        run: |
+          mkdir test_operator_work_dir output_logs
+          podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_success_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
--- a/roles/validate_operator_bundle/tasks/main.yml
+++ b/roles/validate_operator_bundle/tasks/main.yml
@@ -59,7 +59,7 @@
 
     - name: "Fail if Pyxis response is not 200"
       fail:
-        msg: "Error collecting OCP version range from Pyxis.  Error message: {{ supported_indices_json.content }}"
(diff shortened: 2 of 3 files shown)


roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -59,7 +59,7 @@
 
     - name: "Fail if Pyxis response is not 200"
       fail:
-        msg: "Error collecting OCP version range from Pyxis.  Error message: {{ supported_indices_json.content }}"
+        msg: "Error collecting OCP version range from Pyxis - status {{ supported_indices_json.status_code }}.  Error message: {{ supported_indices_json.content }}"
```
lslebodn, line 62: I cannot see code coverage the line. You can easily get code different than 200 
with typo in ocp version.

BTW you need't cover all test-cases in midstream ci.
It be simpler to test in upstream ci.
https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/blob/master/.github/workflows/build_and_test_upstream_ci_image.yml

Because you can "mock" different values of the "com.redhat.openshift.versions" label with modification of the {{ work_dir }}/bundle-skopeo-inspect.json file

Adding more images for different value of the `com.redhat.openshift.versions` does not scale

roles/validate_operator_bundle/tasks/main.yml
```diff
@@ -71,8 +71,8 @@
         msg: "Indices: {{ supported_indices['data'] }}"
 
     - name: "Set support_v4_9 to false if there are no supported versions"
-      set_fact:
-        support_v4_9: false
+      fail:
+        msg: "No supported OCP versions found in Pyxis.  Please verify that the version label is valid: {{ ocp_version_annotation }}"
```
lslebodn, line 75: Cannot see test coverage for the message.
You can get empty list for valid syntax but "ancient" OCP version.

```
> GET /api/containers/v1/operators/indices?sort_by=data.ocp_version&ocp_versions_range==v4.3&organization=redhat-operators HTTP/2
> Host: catalog.redhat.com
> user-agent: curl/7.69.1
> accept: */*
> 
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* Connection state changed (MAX_CONCURRENT_STREAMS == 100)!
< HTTP/2 200 
< server: gunicorn/20.0.4
< content-type: application/json
< content-length: 64
< trace_id: 0xe7db3f8db0cd4be99628e7d605033962
< access-control-allow-origin: *
< cache-control: no-cache, no-store, max-age=0, must-revalidate
< pragma: no-cache
< expires: 0
< x-content-type-options: nosniff
< x-frame-options: DENY
< x-xss-protection: 1 ; mode=block
< referrer-policy: no-referrer
< date: Thu, 16 Sep 2021 14:13:56 GMT
< set-cookie: 377ca4b9aa485619009b75499e9298fd=c2c6d63abc7a07c53af8e6f58ccc83a0; path=/; HttpOnly; Secure
< 
{
  "data": [],
  "page": 0,
  "page_size": 100,
  "total": 0
}
```

lslebodn (changes requested):

roles/validate_operator_bundle/tasks/main.yml
14rcole: I'm a bit confused here.  Isn't that workflow (as well as the `build_and_test_midstream_ci_image.yml` workflow) used to test the CI image itself and not the playbook code?

lslebodn: We do not have any CI image. There is just midstream image which is published.
It is just a way run "tests" in upstream on fedora in github actions

.github/workflows/build_and_test_upstream_ci_image.yml
```diff
@@ -32,3 +32,15 @@ jobs:
...
+        run: |
+          mkdir test_operator_work_dir output_logs
+          podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_success_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
+      
+      - name: "Run the bundle validation test for a bundle with a syntactically incorrect OCP version range --> negative test"
+        continue-on-error: false
+        run: |
+          mkdir test_operator_work_dir output_logs
+          podman run -it -v $PWD:/project/operator-test-playbooks -v ./Dockerfiles/ci/example-metadata-without-alm-annotations:/project/operator_dir -v ./output_logs:/project/output -v ./test_operator_work_dir:/project/test_operator_work_dir -e TEST_NAME=test_for_report_success_empty_alm_examples -e OPERATOR_DIR=/project/operator_dir -e WORK_DIR=/project/output -e OPERATOR_WORK_DIR=/project/test_operator_work_dir/ upstream_ci_image:latest
```
lslebodn, line 46: I am not sure what do you test here.
Cause I cannot se any checking for messages or for return code.

Using github actions as a testing frameworks is a no go.

We already wrote test with python unittest
https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/blob/master/Dockerfiles/ci/run_tests.py

They are called implicitly cause the file is set as CMD in Dockerfile
https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/blob/master/Dockerfiles/ci/Dockerfile#L14

dirgim:
Closing in favor of #249

PR #247, opened 2021-09-17
fix: propagates test script error code correctly

bartoszmajsak:
Right now if test script fails at some point it will log error but exit with `0`. This is because it will take `$?` return code from echo which prints the original error code.

This PR keeps actual error code in the variable to avoid this issue.


bartoszmajsak:
General question about this script - do you plan to adjust it after the repo split?

bartoszmajsak:
Hello? :)

lslebodn:
@J0zi Is upstream directory used from the master branch?
Maybe MR should be created to different branch. 
Similar question as in #244

PR #256, opened 2021-10-05
[CVP-2407] Ensure correct data types in parsed_operator_data.yml

dirgim:
* Add quotes to parsed_operator_data to ensure strings where required


lslebodn (approved):
LGTM

PR #327, opened 2022-04-07
Add check for openshift subscription

nmars:
Added code to parse the `operators.openshift.io/valid-subscription`  annotation from the CSV and add this information to the `parsed_operator_data.yml` artifact.

### DIFF
--- a/roles/parse_operator_bundle/templates/parsed_operator_data.yml.j2
+++ b/roles/parse_operator_bundle/templates/parsed_operator_data.yml.j2
@@ -13,6 +13,14 @@ operator_allnamespaces_support: {{ operator_allnamespaces_support }}
 operator_ownnamespace_support: {{ operator_ownnamespace_support }}
 operator_singlenamespace_support: {{ operator_singlenamespace_support }}
 operator_multinamespace_support: {{ operator_multinamespace_support }}
+{% if operator_valid_subscription|length < 1 %}
+operator_valid_subscription: []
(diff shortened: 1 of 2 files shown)


roles/parse_operator_bundle/templates/parsed_operator_data.yml.j2
```diff
@@ -13,6 +13,7 @@ operator_allnamespaces_support: {{ operator_allnamespaces_support }}
 operator_ownnamespace_support: {{ operator_ownnamespace_support }}
 operator_singlenamespace_support: {{ operator_singlenamespace_support }}
 operator_multinamespace_support: {{ operator_multinamespace_support }}
+operator_valid_subscription: {{ operator_valid_subscription }}
```
lslebodn, line 16: Have you consider omitting `operator_valid_subscription` the same as for `crd_paths` few lines after?
IMHO relying on default value "Not found'" is not ideal.

nmars: Good suggestion

dirgim:
LGTM

dirgim (approved):

roles/parse_operator_bundle/templates/parsed_operator_data.yml.j2
lslebodn: thank you for changes

PR #341, opened 2022-04-27
Parse OpenShift subscription annotation from CSV

nmars:
Added code to parse `operators.openshift.io/valid-subscription` from CSV annotations.

### DIFF
--- a/Dockerfiles/ci/Dockerfile
+++ b/Dockerfiles/ci/Dockerfile
@@ -3,7 +3,7 @@ WORKDIR /project
 ARG OPERATOR_SDK_VERSION=v1.16.0
 RUN export ARCH=$(case $(arch) in x86_64) echo -n amd64 ;; aarch64) echo -n arm64 ;; *) echo -n $(arch) ;; esac);\
     export OS=$(uname | awk '{print tolower($0)}');\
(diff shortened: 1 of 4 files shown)


nmars:
My previous (reverted) pull request had an incorrect filter at line 113 in `roles/parse_operator_bundle/tasks/main.yml`

The previous PR had `| default([])`

So if a value was present in the CSV ansible would parse it as a string instead of a list.

The parsed_operator_data.j2 would then split the string and print something like:

```
operator_valid_subscription:
  - "["
  - """
  - "O"
  - "p"
  - "e"
  - "n"
  - "S"
  - "h"
  - "i"
  - "f"
  - "t"
  - " "
...
```

The current PR corrects this.

lslebodn:
It would be nice to write some tests in upstream.
* https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/blob/master/Dockerfiles/ci/run_tests.py

Ideally for case without required annotation and with them.

nmars:
Tests added.

lslebodn:
> Tests added 

Thank you very much.

dirgim (approved):
LGTM

PR #371, opened 2022-08-10
use `--no-cache-dir` flag to `pip` in dockerfiles to save space

Rajpratik71:
using the "--no-cache-dir" flag in pip install, make sure downloaded packages
by pip don't cache on the system. This is a best practice that makes sure
to fetch from a repo instead of using a local cached one. Further, in the case
of Docker Containers, by restricting caching, we can reduce image size.
In terms of stats, it depends upon the number of python packages
multiplied by their respective size. e.g for heavy packages with a lot
of dependencies it reduces a lot by don't cache pip packages.

Further, more detailed information can be found at

https://medium.com/sciforce/strategies-of-docker-images-optimization-2ca9cc5719b6

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>


upstream/roles/install_iib/templates/Dockerfile-workers.j2
```diff
@@ -75,6 +75,6 @@ RUN sed -i -e 's|^#mount_program|mount_program|g' /etc/containers/storage.conf
 COPY docker/libpod.conf /usr/share/containers/libpod.conf
 
 COPY . .
-RUN pip3 install -r requirements.txt --no-deps --require-hashes
-RUN pip3 install . --no-deps
+RUN pip3 install --no-cache-dir -r requirements.txt --no-deps --require-hashes
+RUN pip3 install --no-cache-dir . --no-deps
```
lslebodn, line 79: You might want to merge two RUn commands to the single one 
```
RUN pip3 install --no-cache-dir -r requirements.txt --no-deps --require-hashes \
    && pip3 install --no-cache-dir . --no-deps
```

Rajpratik71: I think this way is fine, as it is installing dependencies, using "requirements.txt" in the first line then installing the actual package later.

lslebodn: The description of PR mention best practices.
And this is one of best practices.
* https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers

And you you decided to touch lines it would be so such simple change as well.

Rajpratik71: @lslebodn  All operation is happening in a single "RUN" using "&&", so it will not create a new layer.

lslebodn: I can still see two `RUN` instructions in the diff. Maybe you forgot to push changes :-)

Rajpratik71: Sorry @lslebodn my bad, corrected and pushed

lslebodn:
@nmars bump for review

PR #374, opened 2022-09-07
[CVP-3217] Fix issue parsing current channel

nmars:
This ensures that when the current channel is entirely numerical, for example like "4.10", it does not get truncated to "4.1".


14rcole:
lgtm

lslebodn:
LGTM

But please add regression test for better test code coverage into https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/blob/master/Dockerfiles/ci/run_tests.py
Can be done in different PR is users are blocked.

PR #422, opened 2022-12-14
[CVP-3364] Ensure default channel parsed as string

nmars:
- Fixed playbook to ensure that default channel is parsed as a string, the same as was done in #374 for current channel.


samvarankashyap (approved):
lgtm

dirgim (approved):
+1

PR #447, opened 2023-08-14
[CVP-2652] Parse infrastructure feature annotations from CSV

asergienk:


roles/parse_operator_bundle/tasks/main.yml
```diff
@@ -112,6 +112,19 @@
       operator_capabilities: "{{ (csv_vars | from_yaml).metadata.annotations.capabilities }}"
       operator_valid_subscription: "{{ ((csv_vars | from_yaml).metadata.annotations['operators.openshift.io/valid-subscription'] | default('[]')) | from_json }}"
 
+  - name: "Determine and set fact for infrastructure-related annotations"
+    set_fact:
+      operator_feature_disconnected: "{{ (csv_vars | from_yaml).metadata.annotations['features.operators.openshift.io/disconnected'] | default(None) }}"
```
lslebodn, line 117: based on documentation all these new annotations denote support for respective feature in boolean values, so either true or false

will we get boolean type or string

asergienk: Hi @lslebodn That's right, these 10 new annotations could be either true or false. 

Here is the reason I'm setting `default(None)` which is basically just an empty value:
Let's consider this example in the OVP pipeline:
* None of these 10 new annotations defined in the CSV
* so when we try reading them in the CSV and they don't exist there then these values would be defaulted to `default(None)` which is an empty value like `operator_feature_disconnected: ` 
* and then I will check if the values are anything other than true/false I will print a Warning
* so in this example a Warning will be printed

However, let's say I set `default(false)` instead, then in the above situation even if the annotations are not present in the CSV they will be set to `false` which is actually one of the correct values. So it will be confusing to understand whether the value is false because it was set to false by the product team OR is it false because I set is to false by default.

Please let me know if my understanding is completely wrong :D

lslebodn: A) I asked whether the line will set value to boolean **true/false** or to string **"true"/"false"** :-)
B) Do we need to check values to be a boolean or missing as part of "parsing" or they would be validated somewhere else
 People tent to do mistake and one can put "yes/no/maybe" ...

C) I agree that `default(false)` does not make a sense 
the question is if annotation is not defined whether "None/null" should be used as value in the file parsed_operator_data.yml or missing annotation should not be put there at all but that is not important much
both of them might work

asergienk: A) It will be set to boolean true/false
B) At this point our goal is to just check two things:
   * all 10 annotations are present in the CSV 
   * they are set to boolean true/false
If any of the annotations are missing or not true/false then we will print a warning.
C) That made me thinking. Now I'm thinking maybe missing annotations should not be present in parsed_operator_data.yml
I think it could make it more clear

asergienk: @lslebodn thanks for the suggestion! updated this PR with the new changes

asergienk:
One issue was revealed while running a sample operator job:
http://external-ci-coldstorage.datahub.redhat.com/cvp/cvp-redhat-operator-bundle-image-validation-test/e2e-test-operator-bundle-container-8.0-11/6453aeb0-9ef6-4635-984a-14e9d420cf58/operator-metadata-preparation-bundle-image-output.txt

here is a solution: https://github.com/ansible/ansible/issues/74851#issuecomment-852203455
I used `ignore_errors` and it works now

PR #450, opened 2023-08-28
[CVP-3881] Use string for infrastructure feature annotations

asergienk:
1. Changed boolean to string for new infrastructure feature annotations
2. Changed `features.operators.openshift.io/fipsmode` to `features.operators.openshift.io/fips-compliant` as per new requirenments in https://issues.redhat.com/browse/PORTENABLE-525

### DIFF
--- a/Dockerfiles/ci/run_tests.py
+++ b/Dockerfiles/ci/run_tests.py
@@ -274,7 +274,7 @@ class RunOperatorTestPlaybookTests(unittest.TestCase):
             parsed_output = fd.read()
             print(parsed_output)
             self.assertNotIn("operator_feature_disconnected", parsed_output)
-            self.assertNotIn("operator_feature_fipsmode", parsed_output)
+            self.assertNotIn("operator_feature_fips_compliant", parsed_output)
             self.assertNotIn("operator_feature_proxy_aware", parsed_output)
             self.assertNotIn("operator_feature_cnf", parsed_output)
             self.assertNotIn("operator_feature_cni", parsed_output)
@@ -289,7 +289,7 @@ class RunOperatorTestPlaybookTests(unittest.TestCase):
         work_dir = operator_work_dir
         operator_dir = "{}/test-operator".format(operator_work_dir)
         operator_bundle_dir = "{}/operator-bundle".format(operator_work_dir)
-        bundle_image = "quay.io/cvpops/test-operator:with-infrastructure-features"
+        bundle_image = "quay.io/cvpops/test-operator:with-infrastructure-features-as-strings"
         exec_cmd = "ansible-playbook -vvv -i localhost, --connection local \
                     operator-test-playbooks/extract-operator-bundle.yml \
                     -e 'operator_dir={operator_dir}' \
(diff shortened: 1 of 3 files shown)


Dockerfiles/ci/run_tests.py
```diff
@@ -309,16 +309,16 @@ def test_extract_operator_bundle_with_infrastructure_features(self):
...
-            self.assertIn('operator_feature_fipsmode: True', parsed_output)
-            self.assertIn('operator_feature_proxy_aware: False', parsed_output)
-            self.assertIn('operator_feature_cnf: False', parsed_output)
-            self.assertIn('operator_feature_cni: True', parsed_output)
-            self.assertIn('operator_feature_csi: True', parsed_output)
-            self.assertIn('operator_feature_tls_profiles: True', parsed_output)
-            self.assertIn('operator_feature_token_auth_aws: False', parsed_output)
-            self.assertIn('operator_feature_token_auth_azure: False', parsed_output)
-            self.assertIn('operator_feature_token_auth_gcp: False', parsed_output)
+            self.assertIn('operator_feature_disconnected: "False"', parsed_output)
```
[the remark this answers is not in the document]
asergienk: Hi @lslebodn , yes, that's my confusion. When I run a sample operator job everything is lower case:
http://external-ci-coldstorage.datahub.redhat.com/cvp/cvp-redhat-operator-bundle-image-validation-test/e2e-test-operator-bundle-container-8.0-11/3e4859f7-f8dd-4db9-978d-6fe7176af6f3/parsed_operator_data.yml

But somehow here, in this test, the values are being parsed as uppercase "True"/"False", not sure why

asergienk:
Testing:

1. This repo and branch (expected: all annotations missing):
http://external-ci-coldstorage.datahub.redhat.com/cvp/cvp-redhat-operator-bundle-image-validation-test/e2e-test-operator-bundle-container-8.0-11/f4a601bf-f45d-4372-b379-7f197dcbda80/cvp-test-report.html

2. This repo and branch + image:cvpops/test-operator:with-infrastructure-features-as-strings (expected: all annotation defined as "true"/"false"):
http://external-ci-coldstorage.datahub.redhat.com/cvp/cvp-redhat-operator-bundle-image-validation-test/e2e-test-operator-bundle-container-8.0-11/6d7ce0c0-51d5-4901-80b8-3e2cfec329ca/cvp-test-report.html

roles/parse_operator_bundle/templates/parsed_operator_data.yml.j2
```diff
@@ -31,32 +31,32 @@ crd_paths:
 {% endif %}
 ocp_versions: "{{ ocp_versions }}"
 {% if operator_feature_disconnected is defined %}
-operator_feature_disconnected: {{ operator_feature_disconnected }}
+operator_feature_disconnected: "{{ operator_feature_disconnected | lower }}"
```
lslebodn, line 34: can try to use just quotes without `lower`
```
- operator_feature_disconnected: {{ operator_feature_disconnected }}
+ operator_feature_disconnected: "{{ operator_feature_disconnected }}"
```

Otherwise user can pass value `MAYBE` and it would be converted to `maybe`
and that is not really expected

asergienk: without `lower` Ansible parses those values as "True"/"False", this goes back to the discussion in the first thread

asergienk: But somehow if I run a real OVP job then in parsed_operator_data.yml the values are "true"/"false", so maybe I should keep it as is here? 
http://external-ci-coldstorage.datahub.redhat.com/cvp/cvp-redhat-operator-bundle-image-validation-test/e2e-test-operator-bundle-container-8.0-11/3e4859f7-f8dd-4db9-978d-6fe7176af6f3/parsed_operator_data.yml

However, now I see a different issue with using quotes. What if user passes a boolean true, and we convert it to "true" by using the quotes around it which makes it a string

lslebodn: > What if user passes a boolean true, and we convert it to "true" by using the quotes around it which makes it a string

That is tiny disadvantage and should be probably discussed with OCP folks; whether it should fail or not.
But IIUC operator sdk should validate that in the future anyway; and thus CVP would fail a bit later than in parsing phase.

unfortunately, we cannot avoid conversion of bool -> string with current implementation of parse metadata in `ansible`
But IMHO it is a bit smaller problem than automatic lowercasing of strings

BTW one can see that in unit test that bool is converted to string as
`operator_feature_token_auth_azure: "False"\noperator_feature_token_auth_gcp: "False"\n'`
which is not lowercase . So it can be easily detected even though you would not be able to distinguish between
```
features.operators.openshift.io/disconnected: true
vs
features.operators.openshift.io/disconnected: "True"
```

asergienk: Yes, I agree that using `lower` is definitely not a good idea. In this case, I'll return to my first implementation where I have to use "True"/"False" in the test. Seems like it is the best option.
Thank you!

lslebodn: `"True"/"False"` 
That would be for boolean. Which is OK for negative test case.
It should be just clear from test either by adding comments or jsut more self descriptive test name
(or combination of both :-)

PR #451, opened 2023-09-04
Copy sql database to the fbc index

Allda:
This change makes a current community index image compatible with IIB service by extracting the SQL database and pushing it to a hidden location used by IIB.

The community indexes are build with SQL-based database and later converted to fbc for ocp >=4.11 (using `opm migrate`). This solution bridges a gap between SQL-based and fbc catalogs to allow the use of the same opm API for both.

### DIFF
--- a/upstream/roles/produce_fbc/tasks/main.yml
+++ b/upstream/roles/produce_fbc/tasks/main.yml
@@ -50,6 +50,15 @@
     loop_var: rf_render_target
   when: pfbc_render_from == "index"
 
+- name: "Extract /database/index.db from the index"
+  include_role:
+    name: extract_files_from_image
+  vars:
+    effi_image: "{{ pfbc_sql_index }}"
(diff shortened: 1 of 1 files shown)


upstream/roles/produce_fbc/tasks/main.yml
```diff
@@ -65,6 +74,13 @@
     chdir: "{{ pfbc_extracted_fbc_location }}"
   register: pfbc_rc
 
+- name: Copy /database/index.db to a hidden IIB directory
```
mantomas, line 77: Name should be quoted here I think, at least for consistency.

mporrato (approved):

upstream/roles/produce_fbc/tasks/main.yml
Allda: jaj I clicked on merge before noticed this comment. I'll fix it in a separate commit.

PR #452, opened 2023-09-06
Add 4.15 version to excluded list

Allda:


upstream/roles/check_image_availability/tasks/main.yml
```diff
@@ -108,7 +108,7 @@
 
       when:
         - not run_upstream|bool
-        - current_openshift_run is defined and current_openshift_run not in ["4.12", "4.13", "4.14"]
+        - current_openshift_run is defined and current_openshift_run not in ["4.12", "4.13", "4.14", "4.15"]
```
mporrato, line 111: I think we can just do this:
```suggestion
        - current_openshift_run is defined and current_openshift_run is version(current_openshift_run, "4.12", "<")
```
and forget about it for future index releases.

mporrato (approved):

upstream/roles/check_image_availability/tasks/main.yml
Allda: I changed it as you suggested. Thanks!

PR #453, opened 2023-09-12
[CVP-3890] Do string comparison of channels.v1 label

nmars:
Do string comparison of `operators.operatorframework.io.bundle.channels.v1` label from `annotations.yaml`.

### DIFF
--- a/roles/parse_operator_bundle/tasks/bundle_sanity_checks.yml
+++ b/roles/parse_operator_bundle/tasks/bundle_sanity_checks.yml
@@ -29,7 +29,7 @@
 - name: "Check if the operators.operatorframework.io.bundle.channels.v1 from annotation.yaml matches the bundle image label"
   fail:
     msg: "The operators.operators.operatorframework.io.bundle.channels.v1 value in the annotations yaml doesn't match the corresponding bundle image label!"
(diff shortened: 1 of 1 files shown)


roles/parse_operator_bundle/tasks/bundle_sanity_checks.yml
```diff
@@ -29,7 +29,7 @@
 - name: "Check if the operators.operatorframework.io.bundle.channels.v1 from annotation.yaml matches the bundle image label"
   fail:
     msg: "The operators.operators.operatorframework.io.bundle.channels.v1 value in the annotations yaml doesn't match the corresponding bundle image label!"
-  when: annotations_vars.annotations['operators.operatorframework.io.bundle.channels.v1'] != skopeo_inspect_json.Labels['operators.operatorframework.io.bundle.channels.v1']
```
lslebodn, line 32: container image already returns string even for `LABEL operators.operatorframework.io.bundle.channels.v1=3.11`

so `| string` is not needed for `skopeo_inspect_json`

and string seems to be expected for the annotation
https://github.com/operator-framework/operator-registry/blob/5ca4a9f7dcb7be737b18f952f72a3a4dc195d841/pkg/registry/types.go#L157C2-L157C10

PR #455, opened 2023-10-13
[ISV-4161] Ensure target bundle exists before upgrading

mporrato:


ezopezo (approved):

upstream/roles/test_operator_update/tasks/main.yml
```diff
@@ -89,6 +91,16 @@
     - name: "Create namespace for upgrade tests"
       shell: "{{ oc_bin_path }} create ns {{ tou_namespace_for_upgrades }}"
 
+    - name: "Check if {{ tou_operator_bundle_for_upgrade }} exists"
+      shell: "skopeo inspect docker://{{ tou_operator_bundle_for_upgrade }}"
```
lslebodn, line 95: a bit of nitpick.
`--no-tags` migth be useful here.
```
       --no-tags, -n

       Do not list the available tags from the repository in the output.  When
       true,  the  RepoTags array will be empty.  Defaults to false, which in‐
       cludes all available tags.
```

It is much faster especially when inspected image has many (100+) tags in the repository.

PR #456, opened 2023-12-03
[CVP-3949] Add olm.skipRange and olm.replaces to parsed_operator_data

yashvardhannanavati:


yashvardhannanavati:
@nmars @samvarankashyap Could you please take a look?

lslebodn:
@yashvardhannanavati It would be nice to write some tests
* https://github.com/redhat-openshift-ecosystem/operator-test-playbooks/blob/master/Dockerfiles/ci/run_tests.py

ideally positive and negative (e.g. with missing olm.replaces in metadata)

PR #458, opened 2024-02-12
[CVP-3985] add all targeted channels data to parsed_operator_data.yml

yashvardhannanavati:


yashvardhannanavati:
@asergienk @nmars @samvarankashyap Could you PTAL?

PR #464, opened 2024-08-04
[ISV-5091] Optimize container tool storage configuration

mavaras:
**TLDR**: specifying a different storage driver (overlay) from the default one (VFS) fixed the issue as it is more optimized for storage.

- Adds a new Ansible step controlled by the boolean variable `optimize_build_storage` to apply a storage configuration that saves disk space and make podman build execution safer. Take into account that the podman instance used here is NOT the same as the one we handle in community-operators-pipeline. The only way to customize it was to edit it straight into the Ansible config that invokes it.
- Adds extra parameters variable to podman build command. This allows to easily add new configuration options and for debugging just editing [this list](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline/blob/e1de0e6d6ebddc0f4fc6f7f4cec2a6f6cd57c926/ci/scripts/opp.sh#L45). The real fix to the issue is to setup overlay as storage driver with `--storage-driver`.

Example start-to-end pipeline run with this changes being applied successfully: https://github.com/k8s-operatorhub/community-operators/actions/runs/10237309646/job/28322358625

Some references:
- https://github.com/containers/buildah/issues/4423
- https://github.com/containers/buildah/issues/2850

### DIFF
--- a/upstream/roles/build_app_registry/defaults/main.yml
+++ b/upstream/roles/build_app_registry/defaults/main.yml
@@ -3,6 +3,7 @@ testing_bin_path: "{{ work_dir }}/bin"
 opm_bin_path: "{{ testing_bin_path }}/opm"
 bundle_export_dir_app_registry: "{{ work_dir }}/app_registry/upstream-community-operators"
 container_tool: "docker"
+container_tool_extra_params: ""
 opm_container_tool: "docker"
 opm_container_tool_index: "docker"
 container_push_extra: ""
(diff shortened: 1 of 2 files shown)


Allda (approved):

mporrato (approved):

acornett21:
Is the plan after this is merged to re-run every failed job (in order), and also some jobs that show as passed, but the ohio job didn't run?

acornett21:
What about jobs like this that show green but the entire pipeline never ran? Is there a way to track these and make sure they are also re-ran?

- https://github.com/k8s-operatorhub/community-operators/actions/runs/10221231385

mavaras:
> What about jobs like this that show green but the entire pipeline never ran? Is there a way to track these and make sure they are also re-ran?
> 
> * https://github.com/k8s-operatorhub/community-operators/actions/runs/10221231385

Yes, I'll be checking all of these since the issue started.

PR #478, opened 2026-06-24
fix(ISV-7300): remove the need for authentication in test pipeline

BorekZnovustvoritel:
Assisted-by: Claude-4.6-Opus (Cursor)

### DIFF
--- a/roles/install_operator_prereqs/tasks/main.yml
+++ b/roles/install_operator_prereqs/tasks/main.yml
@@ -31,7 +31,7 @@
 
 - name: "Download and extract go binary to {{ testing_bin_path }}"
   unarchive:
-    src: "https://storage.googleapis.com/golang/go{{ go_version }}.linux-amd64.tar.gz"
+    src: "https://dl.google.com/go/go{{ go_version }}.linux-amd64.tar.gz"
     dest: "{{ testing_bin_path }}"
     remote_src: yes
   tags:
--- a/upstream/ansible.cfg
+++ b/upstream/ansible.cfg
@@ -2,7 +2,8 @@
 roles_path = ../roles
 display_skipped_hosts = no
 display_ok_hosts = yes
-stdout_callback = yaml
+stdout_callback = ansible.builtin.default
+result_format = yaml
 callback_whitelist = log_plays
 
 [callback_log_plays]
--- a/upstream/roles/multiarch_fbc/tasks/process_archs.yml
+++ b/upstream/roles/multiarch_fbc/tasks/process_archs.yml
@@ -1,7 +1,6 @@
 ---
 - name: "Pull the base image {{ container_tool }} pull --arch={{ mfbc_architecture }} {{ mirror_multiarch_image }}"
   shell: "{{ container_tool }} pull --arch={{ mfbc_architecture }} {{ mirror_multiarch_image }}"
-  when: registry_redhat_io_user is defined and registry_redhat_io_token is defined
 
 - name: "Retag the base image"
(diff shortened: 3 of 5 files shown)


roles/install_operator_prereqs/tasks/main.yml
```diff
@@ -31,7 +31,7 @@
 
 - name: "Download and extract go binary to {{ testing_bin_path }}"
   unarchive:
-    src: "https://storage.googleapis.com/golang/go{{ go_version }}.linux-amd64.tar.gz"
+    src: "https://dl.google.com/go/go{{ go_version }}.linux-amd64.tar.gz"
```
BorekZnovustvoritel, line 34: This URL changed (the original no longer works, the image was not buildable).

upstream/roles/multiarch_fbc/tasks/process_archs.yml
```diff
@@ -1,7 +1,6 @@
 ---
 - name: "Pull the base image {{ container_tool }} pull --arch={{ mfbc_architecture }} {{ mirror_multiarch_image }}"
   shell: "{{ container_tool }} pull --arch={{ mfbc_architecture }} {{ mirror_multiarch_image }}"
-  when: registry_redhat_io_user is defined and registry_redhat_io_token is defined
```
BorekZnovustvoritel, line 4: This condition did not make sense since the image is retagged in the next step unconditionally. If the user and/or token were not defined, this playbook just failed.

upstream/roles/produce_fbc/tasks/main.yml
```diff
@@ -105,28 +105,8 @@
...
-    - name: "Generate cache locally"
-      # shell: "{{ opm_bin_path }} serve {{ pfbc_fbc_config_path }} --cache-dir={{ pfbc_local_cache_dir }} --cache-only --termination-log /dev/null"
-      shell: "{{ container_tool }} run -it --rm --entrypoint /bin/opm -v {{ pfbc_extracted_fbc_location }}:/conf -u 1001:0 {{ mirror_multiarch_image }} serve /conf/configs --cache-dir=/conf/cache --cache-only --termination-log /dev/null"
-
-    # - name: "Copy cache"
-    #   synchronize:
-    #     src: "{{ pfbc_local_cache_dir }}/"
-    #     dest: "{{ pfbc_extracted_fbc_location }}/cache/"
+    - name: "Generate cache locally using host opm to solve arch issues"
+      shell: "{{ opm_bin_path }} serve {{ pfbc_fbc_config_path }} --cache-dir={{ pfbc_extracted_fbc_location }}/cache --cache-only --termination-log /dev/null"
```
BorekZnovustvoritel, line 109: I found out that the custom cache IS needed. It's because without it, the playbook would try to run opm from within the multuarch image. It passed for `amd64`, but other arches failed immediately because the `opm` binary was not runnable.

However the cache can be created without ever reaching out to `registry.redhat.io`, so the credential requirement can be removed from here.

upstream/ansible.cfg
```diff
@@ -2,7 +2,8 @@
 roles_path = ../roles
 display_skipped_hosts = no
 display_ok_hosts = yes
-stdout_callback = yaml
+stdout_callback = ansible.builtin.default
+result_format = yaml
```
BorekZnovustvoritel, line 6: These were missed previously. `stdout_callback = yaml` has been removed from Ansible and the image was not buildable without this change.

Allda (approved):

RichardPlesnik (approved):

PR #481, opened 2026-07-13
fix(ISV-7433): filter out non-arch artifacts in registry

mantomas:
This change is needed to support transition to a new version of OPM - architectures are decided based on all artifacts in the repository [like this](https://github.com/k8s-operatorhub/community-operators/actions/runs/29209144008/job/86693972307#step:3:4763) and fail on non-arch items like SBOMs or attestations with `unknown` arch (failure example [here](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline/actions/runs/26947797095/job/79505268770#step:3:3264)). This PR filters out unknown and empty arches.

Only the second change in mirror_single.yml is actually needed, the first change is kept for parity in logic.

Tested locally with newest OPM using updated operator-test-playbooks.

Assisted-by: Cursor

### DIFF
--- a/upstream/roles/multiarch_fbc/tasks/main.yml
+++ b/upstream/roles/multiarch_fbc/tasks/main.yml
@@ -52,13 +52,6 @@
   set_fact:
     multiarch_fbc_architectures: "{{ mfbc_archs_inspect_result.stdout_lines | list }}"
 
-- name: "Filter unsupported architectures"
-  set_fact:
-    multiarch_fbc_architectures: >-
-      {{ multiarch_fbc_architectures
-         | reject('equalto', '')
-         | list }}
-
 - name: "Print archs list'"
   debug:
     var: multiarch_fbc_architectures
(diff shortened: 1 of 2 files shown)


upstream/roles/multiarch_fbc/tasks/main.yml
```diff
@@ -42,13 +42,23 @@
...
 
 - name: "Parse archs to list"
   set_fact:
     multiarch_fbc_architectures: "{{ mfbc_archs_inspect_result.stdout_lines | list }}"
 
+- name: "Filter unsupported architectures"
+  set_fact:
+    multiarch_fbc_architectures: >-
+      {{ multiarch_fbc_architectures
+         | reject('equalto', '')
```
BorekZnovustvoritel: I wonder why this is needed only in this instance of calling the same code in multiple places 🤔

BorekZnovustvoritel (approved):
Seems like this will work, I just have a single non-blocking question.

upstream/roles/multiarch_fbc/tasks/main.yml
mantomas: Valid point - I actually decided to drop the additional task. The agent reasoning was somehow unclear so I dig deeper and it would be really rare case if empty string is passed as architecture (not valid by OCI standard) - null or empty is common and it is already filtered in the jq.

PR #483, opened 2026-08-20
fix(ISV-7506): upgrade to F44 on all dockerfiles

bclindner:
- **fix(ISV-7506): upgrade to F44 on all dockerfiles**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
-FROM fedora:37
-RUN dnf install -y git podman buildah crun-0:1.6-2.fc37 python3-libselinux python3-pip rsync python3-dnf-plugin-versionlock
-RUN dnf versionlock crun-*
-RUN pip3 install ansible==6.5.0 jmespath ansible-runner
+FROM quay.io/fedora/fedora:44
+RUN dnf install -y git-core podman buildah python3-pip rsync
+RUN pip3 install setuptools ansible==6.5.0 jmespath ansible-runner
 RUN mkdir -p /playbooks
 COPY roles/ /playbooks/roles/
 COPY filter_plugins/ /playbooks/filter_plugins/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM quay.io/fedora/fedora:44
 RUN dnf install -y git-core podman buildah python3-pip rsync
-RUN pip3 install setuptools ansible==6.5.0 jmespath ansible-runner
+RUN pip3 install setuptools ansible==14.3.1 jmespath ansible-runner
 RUN mkdir -p /playbooks
 COPY roles/ /playbooks/roles/
 COPY filter_plugins/ /playbooks/filter_plugins/
--- a/upstream/ansible.cfg
+++ b/upstream/ansible.cfg
@@ -4,7 +4,7 @@ display_skipped_hosts = no
 display_ok_hosts = yes
 stdout_callback = ansible.builtin.default
 result_format = yaml
-callback_whitelist = log_plays
+callbacks_enabled = community.general.log_plays
 
 [callback_log_plays]
 log_folder = /tmp/ansible_logs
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -1,7 +1,6 @@
-FROM fedora:37
-RUN dnf install -y git podman buildah crun-0:1.6-2.fc37 python3-libselinux python3-pip rsync python3-dnf-plugin-versionlock
-RUN dnf versionlock crun-*
-RUN pip3 install ansible==6.5.0 jmespath ansible-runner
+FROM fedora:44
+RUN dnf install -y git podman buildah python3-pip rsync
```
lslebodn: I would use git-core which will pull less dependencies

Dockerfile
```diff
@@ -1,7 +1,6 @@
-FROM fedora:37
-RUN dnf install -y git podman buildah crun-0:1.6-2.fc37 python3-libselinux python3-pip rsync python3-dnf-plugin-versionlock
-RUN dnf versionlock crun-*
-RUN pip3 install ansible==6.5.0 jmespath ansible-runner
+FROM fedora:44
```
lslebodn: please use either registry.fedoraproject.org/fedora or quay.io/fedora/fedora Just to avoid hitting docker hub and rate limiting

bclindner: Thanks for the callout - fixed both issues

BorekZnovustvoritel:
Are you able to build this locally? I receive this problem:

```text
STEP 11/14: RUN ansible-galaxy collection install ansible.utils community.general
Starting galaxy collection install process
Process install dependency map
[WARNING]: Skipping Galaxy server https://galaxy.ansible.com. Got an unexpected
error when getting available versions of collection ansible.utils: Unknown
error when attempting to call Galaxy at 'https://galaxy.ansible.com/api/':
'CustomHTTPSConnection' object has no attribute 'cert_file'.
'CustomHTTPSConnection' object has no attribute 'cert_file'
ERROR! Unknown error when attempting to call Galaxy at 'https://galaxy.ansible.com/api/': 'CustomHTTPSConnection' object has no attribute 'cert_file'. 'CustomHTTPSConnection' object has no attribute 'cert_file'
Error: building at STEP "RUN ansible-galaxy collection install ansible.utils community.general": while running runtime: exit status 1
```

If that's just on my side, it can be just some CDN hiccup. Otherwise we should take a look into this. Unfortunately the target branch `upstream-community-dev` does not have an automated CI/CD, so stuff like this has to be checked manually.

If you can replicate the error (I just ran `podman build .`, which worked in the previous state), please try to fix it and then you can push that built image to `quay.io/operator_testing/operator-test-playbooks:dev`. After that you can retrigger testing on https://github.com/redhat-openshift-ecosystem/community-operators-pipeline/pull/482 (don't hesitate to push to that branch, it's just for testing) and if it passes, we have most likely solved this.

bclindner:
Good catch, it's the Ansible version. I assumed it was just a local build issue since where maybe it needed some special certs, but no, 6.5.0 is about 4 years old now and it's not a fan of the environment it's running in. I'm seeing what it'll take to get us upgraded to 14.1.0.

mantomas (approved):
LGTM - I was able to build image locally with the latest changes (and the test pipeline looks OK).

Dockerfile
```diff
@@ -1,7 +1,6 @@
-FROM fedora:37
-RUN dnf install -y git podman buildah crun-0:1.6-2.fc37 python3-libselinux python3-pip rsync python3-dnf-plugin-versionlock
-RUN dnf versionlock crun-*
-RUN pip3 install ansible==6.5.0 jmespath ansible-runner
```
BorekZnovustvoritel, line 4: Can you please provide some short commentary why this needed to be changed? I am not implying it is wrong, I just don't quite understand why CRUN was pinned to a specific version and now is not explicitly present at all.

BorekZnovustvoritel (approved):
Just a minor comment, looks good though.

Dockerfile
lslebodn: One would need to check git blame
```
commit 9d696b6dd133323374908f326a23314dc2763c64
Author: Maurizio Porrato <mporrato@redhat.com>
Date:   Fri Jun 9 13:07:21 2023 +0100

    Temporary fix for ISV-3709

```
And "temporary fix" lasted 3 years :-)

bclindner: Yep, Lukas beat me to it- some older versions of crun were preventing us from standing up a specific container image. Unsure of exact details of _why_ newer versions fix it, but I determined we no longer needed after testing w/o it. I'll monitor after this PR is merged to see if we somehow still need it but I get the feeling newer versions have probably fixed the bug for good

PR #487, opened 2026-09-04
fix(ISV-7652): Fix issues after ansible upgrade

RichardPlesnik:


upstream/roles/handle_failed_operators/tasks/main.yml
```diff
@@ -5,7 +5,7 @@
 
 - name: "Adding bundle_failed from 'bundle_index_add_failed' without duplicates"
   set_fact:
-    bundle_failed: "{{ bundle_failed | default([]) }} + {{ bundle_index_add_failed | unique }}"
+    bundle_failed: "{{ (bundle_failed | default([])) + (bundle_index_add_failed | unique | list) }}"
```
lslebodn: Using the union filter directly performs set addition and deduplication in a single step, rendering the subsequent "Removing duplicates from 'bundle_failed'" task redundant:

```suggestion
    bundle_failed: "{{ (bundle_failed | default([])) | union(bundle_index_add_failed | default([])) }}"}}"
```

BorekZnovustvoritel (approved):
Let's try this!

upstream/roles/operator_info/tasks/main.yml
```diff
@@ -140,7 +140,7 @@
             oi_auto_label_error_msg: |
               Automatic cluster versions lables are needed and 'automatic_cluster_version_label' is set to False.
```
lslebodn: Unrelated to your PR but there is a typo in error message `lables` :-)

upstream/roles/handle_failed_operators/tasks/main.yml
BorekZnovustvoritel: Fair, but there are many more things to be fixed/simplified/removed. At the moment, we just want it to work again on a newer Ansible version :sweat_smile: In the future we can try to think of ways how to break out of the grasp of this repo :smile:

upstream/roles/operator_info/tasks/main.yml
RichardPlesnik: fixed :smile:

upstream/roles/handle_failed_operators/tasks/main.yml
RichardPlesnik: Good point, fixed :+1:

JakubDurkac (approved):

upstream/roles/handle_failed_operators/tasks/main.yml
lslebodn: > In the future we can try to think of ways how to break out of the grasp of this repo 😄

let the AI agent rewrite ansibel tasks to the reasonable programing language? :-)

Commetns were just a suggestions feel free to ignore && resolve threads.

lslebodn (approved):
LGTM

upstream/roles/handle_failed_operators/tasks/main.yml
BorekZnovustvoritel: > let the AI agent rewrite ansibel tasks to the reasonable programing language?

Well yes, but actually we already have [operator pipelines](https://github.com/redhat-openshift-ecosystem/operator-pipelines) at home :smile:

If there was a way how to convince everyone to migrate the k8s framework to Operator Pipelines and then actually do the work, then we can get rid of this nonsense completely. Or to Konflux.

PR #489, opened 2026-09-10
ISV-7511: add checksum verification to binary downloads

bclindner:
Adds sha256 digest verification for most utilities downloaded in playbooks. Currently have just about everything using get_url covered, still looking for the best way to do verification for other utilities downloaded via curl etc.


upstream/roles/check_image_availability/tasks/main.yml
lslebodn, line 29: +100 for adding checksum
but maybe you can remove this block in the upstream branch :-)

FYI `yq` is already packaged in fedora as rpm. Feel free to resolve the comment.

tomasbakk:
I'll review this today EOD

ezopezo:
Sorry for late review! I am bit confused, is this pointing out to correct ticket? This seems to me more like ISV-7511 effort.

upstream/roles/install_operator_prereqs/defaults/main.yml
```diff
@@ -5,10 +5,15 @@ oc_version: 4.3.5
 operator_courier_version: 2.1.7
 go_version: 1.13.7
```
ezopezo, line 10: Some downloaded tools (also grpcurl and dyff) still lack checksum validation. Why?

upstream/roles/install_operator_prereqs/defaults/main.yml
```diff
@@ -5,10 +5,15 @@ oc_version: 4.3.5
...
 yq_version: 2.2.1
+yq_checksum: "sha256:54443f0e9796dfb8423a6e33b8175a0f77b8bb0eb55db820ecb84098db2c0ed4"
 umoci_version: v0.4.5
+umoci_checksum: "sha256:1d9d3be0abc7a9fc4eb7d522e73794c561b740b1386ab7febaf69d58092b7af6"
 opm_version: v1.14.2
+opm_checksum: "sha256:6df05bff3d7237986328ee7e50ad74da6cca68fbcdfcf614d50e0166c9f106d6"
 k8s_community_bundle_validator_version: v0.1.0
+k8s_community_bundle_validator_checksum: "sha256:1441717fd7301dadfbf455468c71d974b553876b26c2de1ebe24057b683cd797"
 grpcurl_version: 1.8.6
 dyff_version: 1.5.1
```
ezopezo: Same for these

upstream/local.yml
ezopezo, line 62: Same here (+operator_courier should be probably hash-pinned)

tomasbakk:
I have basically the same question as ezopezo, regarding the missing shas.

upstream/local.yml
lslebodn: operator courier was used for  AppRegistry packaging format of operators IIRC before openshift 4.5)
And the project is archived https://github.com/operator-framework/operator-courier

It cannot work for operator bundles.

 IMHo you should consider to remove it insted of pinning.
 
 my 2 cents.

bclindner:
> Sorry for late review! I am bit confused, is this pointing out to correct ticket? This seems to me more like ISV-7511 effort.

Oh whoops! I am getting my tickets mixed up due to the sheer volume of them, thanks for the catch.

Will leave the branch as it is for now, suspect recreating the PR might not be worth the trouble

upstream/roles/install_operator_prereqs/defaults/main.yml
bclindner: TIL we download files like three different ways... that's my bad, still getting used to Ansible & thought covering all the `get_url` references were enough w/o checking the rest of the vars. Let me go through and get all the others now

upstream/local.yml
bclindner: operator_courier removal is going to blow the scope of this PR wayyyy up, so for now I am opting to just pin it & we're gonna get a ticket on the board for removing the old appregistry packaging flow

bclindner: OK pinning that was a gigantic pain but it works - we have to pull the wheel directly from pypi and verify it that way since the hashes for its dependencies aren't pinned

upstream/roles/install_operator_prereqs/defaults/main.yml
```diff
@@ -1,14 +1,24 @@
...
```
ezopezo, line 24: What about those grpcurl and dyff?

bclindner: You are way better at finding these things than I am clearly... fixed again, gave everything a second pass, didn't see anything outside of `oc` that would be missing (& that's a special case where the version number is determined at runtime)

bclindner: Sorry about the mess on this PR, this is like my third time working with Ansible and my first time making sweeping changes to a playbook repo this big

upstream/roles/install_operator_prereqs/tasks/main.yml
kosciCZ: This is unpinned.

kosciCZ (changes requested):
Additionally, I let codex do a pass to see if anything is unpinned and apart from my other comment it pointed at this, which I think is valid:

> upstream/roles/install_operator_prereqs/tasks/opm_special.yml:7 pulls an opm container image and copies the executable from it without verification. The
>     configured default is mutable quay.io/operator-framework/opm:master

upstream/roles/install_operator_prereqs/tasks/main.yml
bclindner: I was trying to keep it to binaries just to avoid blowing up the scope any further, but if this is the only thing we have left that's not pinned I can set that to the latest commit

bclindner: Done - probably helps this repo hasn't had an update in 7 years (worrying)

